PR #2, opened 2021-09-06
Chat.php - Avy's in Chat

MiteyDrill:
User Avy's should pop up to right of their name/village

Unsure if this will work in all cases / or for all viewing devices - CSS has hard coded values such as 'margin: 10% 0px'; etc

image

### DIFF
--- a/chat.php
+++ b/chat.php
@@ -159,7 +159,7 @@ function chat() {
 	echo "<table class='table' style='width:98%;'>
 		<tr>
 			<th style='width:30%;'>Users</th>
-			<th style='width:66%;'>Message</th>
+			<th style='width:60%;'>Message</th>
 			<th style='width:9%;'>Time</th>
 		</tr>";
 		if(! $system->db_num_rows) {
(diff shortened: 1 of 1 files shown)


MiteyDrill:
Won't work with Staff Tags, might have to make some edits to the css later

chat.php
```diff
@@ -155,23 +155,37 @@ function chat() {
...
 			$purchasers = $system->query("SELECT `premium_credits_purchased` FROM `users` WHERE `user_name` = '$post[user_name]'");
 			$creditsPurchased = $system->db_fetch();
+
+			/*Cextra Zone*/
+			$avyQuery = $system->query("SELECT `avatar_link` FROM `users` WHERE `user_name` = '$post[user_name]'");
+			$avyLink = $system->db_fetch($avyQuery);
 			echo "
 				<tr>
 					<td style='text-align:center;'>
+					<div id=\"user_data_container style='display=block'\">
```
levimeahan, line 178:
```suggestion
					<div id='user_data_container' style='display=block'>
```

levimeahan (approved):
Nice, thanks for doing this!

PR #18, opened 2021-09-10
ak purchased/exp viewer

Algharna:
https://github.com/Algharna/shinobi-chronicles/blob/main/members.php
https://github.com/Algharna/shinobi-chronicles/blob/main/profile.php

Redid it because I had no idea what happened 🤣


levimeahan (commented):
Looks like some merge conflicts, recommend using rebase instead of deleting and recreating the profile.php file.

Can you delete the stat icon images from your fork? That should clear them from this PR, I can't delete them from this PR manually

PR #27, opened 2021-09-12
Update chat and ramen shop

Badoyz:
Fixed warnings for keys post, delete, and min


levimeahan (approved):

levimeahan:
Thanks for fixing these!

PR #29, opened 2021-09-12
Mobile CSS - Menu Update

MiteyDrill:
- added script file
 - changes some html placements
 - AK page menu missing

### DIFF
--- a/arena.php
+++ b/arena.php
@@ -206,18 +206,7 @@ function arena() {
 		$fight_timer = 20;
 		if($_GET['fight']) {
 			if($player->last_ai > time() - $fight_timer) {
-
-				$temp_time = $player->last_ai - (time() - $fight_timer);
-
-				//can't seem to squeeze a <div> around $temp_time
-				//for some reason if there's a tag at the end
-				//it creates another empty <p> tag after the notification message
-				//confused at how and why this happens
-				$temp_msg = <<<HTML
-					<p>Please wait {$temp_time} more seconds!
-				HTML;
-
-				$system->message($temp_msg);
+				$system->message("Please wait " . ($player->last_ai - (time() - $fight_timer)) . " more seconds!");
 			}
 			else if(isset($ai_opponents[$_GET['fight']])) {
 				$_SESSION['ai_id'] = $_GET['fight'];
--- a/arena.php
+++ b/arena.php
@@ -206,7 +206,18 @@ function arena() {
 		$fight_timer = 20;
 		if($_GET['fight']) {
 			if($player->last_ai > time() - $fight_timer) {
-				$system->message("Please wait " . ($player->last_ai - (time() - $fight_timer)) . " more seconds!");
+
+				$temp_time = $player->last_ai - (time() - $fight_timer);
+
+				//can't seem to squeeze a <div> around $temp_time
+				//for some reason if there's a tag at the end
+				//it creates another empty <p> tag after the notification message
+				//confused at how and why this happens
+				$temp_msg = <<<HTML
+					<p>Please wait {$temp_time} more seconds!
+				HTML;
+
+				$system->message($temp_msg);
 			}
 			else if(isset($ai_opponents[$_GET['fight']])) {
 				$_SESSION['ai_id'] = $_GET['fight'];
--- a/chat.php
+++ b/chat.php
@@ -197,7 +197,7 @@ function chat() {
 			echo "
 				<tr>
 					<td style='text-align:center;'>
-					<div id='user_data_container'>
+					<div id='user_data_container' style='display: flex;flex-direction:row'>
 					    <div style='flex-shrink:0;'>
 					        <img style='max-height:45px;max-width:45px;' src='{$userData['avatar_link']}' />
                         </div>
--- a/chat.php
+++ b/chat.php
@@ -203,7 +203,7 @@ function chat() {
                         </div>
 						<div style='display:block;flex-grow:1;'>
 							<a style='display:inline-block;' href='{$system->links['members']}&user={$post['user_name']}' class='$class $statusType'>{$post['user_name']}</a><br />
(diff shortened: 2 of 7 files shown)


levimeahan:
Ah looks like this got the hotkey changes pulled in. I'm okay reviewing them together, but if you want to keep them separate in the future may need to create separate branches for each feature you work on

MiteyDrill:
Okay, I didn't know the older PR would include changes after I submitted the request but it makes sense now. I'll try doing the branch change in the future. Might try the <div> thing later, gonna take a long break. Thanks Lsm:)!

MiteyDrill:
I made some changes that should fix some issues I forgot to check for earlier

MiteyDrill:
image
simple change fixed the buttons

MiteyDrill:
@levimeahan I think I have a simple version of hotkeys set up here, it's hidden in the MobileCSS PR I can rename the PR if it helps

chat.php
```diff
@@ -197,13 +197,13 @@ function chat() {
 			echo "
 				<tr>
 					<td style='text-align:center;'>
-					<div id='user_data_container' style='display: flex;flex-direction:row'>
+					<div id='user_data_container'>
```
levimeahan, line 200: I think this is going to break it, why do we need to change this?

classes/system.php
```diff
@@ -226,7 +226,13 @@ public function message($message, $force_message = false) {
...
     public function printMessage($force_display = false) {
         if(strlen($this->message) && (!$this->message_displayed || $force_display)) {
-            echo "<p class='systemMessage'>$this->message</p>";
+
+            $temp_msg = <<<HTML
+            <p class='systemMessage'> {$this->message} </p>
+            HTML;
+
+            echo $temp_msg;
+
```
levimeahan, line 235:
```suggestion
echo "<p class='systemMessage'>{$this->message}</p>";
```

arena.php
```diff
@@ -206,7 +206,18 @@ function arena() {
...
+
+				//can't seem to squeeze a <div> around $temp_time
+				//for some reason if there's a tag at the end
+				//it creates another empty <p> tag after the notification message
+				//confused at how and why this happens
+				$temp_msg = <<<HTML
+					<p>Please wait {$temp_time} more seconds!
+				HTML;
+
+				$system->message($temp_msg);
```
levimeahan, line 220:
```suggestion
				$system->message("Please wait " . ($player->last_ai - (time() - $fight_timer)) . " more seconds!");
```
Let's sort this out in another PR

classes/system.php
```diff
@@ -226,7 +226,7 @@ public function message($message, $force_message = false) {
     */
     public function printMessage($force_display = false) {
         if(strlen($this->message) && (!$this->message_displayed || $force_display)) {
-            echo "<p class='systemMessage'>$this->message</p>";
+echo "<p class='systemMessage'>{$this->message}</p>";
```
levimeahan, line 229:
```suggestion
            echo "<p class='systemMessage'>{$this->message}</p>";
```

chat.php
MiteyDrill: Oh yeah sorry, I was testing this on the mobile view. Don't change this inline. mb

chat.php
```diff
@@ -197,13 +197,13 @@ function chat() {
...
 					    <div style='flex-shrink:0;'>
 					        <img style='max-height:45px;max-width:45px;' src='{$userData['avatar_link']}' />
                         </div>
-						<div style='display:block;flex-grow:1;'>
-							<a href='{$system->links['members']}&user={$post['user_name']}' class='$class $statusType'>{$post['user_name']}</a><br />
-							<p style='margin: 1px 0 3px;'>
+						<div style='display:inline-block;flex-grow:1;'>
+							<a style='display:inline-block;' href='{$system->links['members']}&user={$post['user_name']}' class='$class $statusType'>{$post['user_name']}</a><br />
```
MiteyDrill, line 205: I think this line will also affect chat

chat.php
levimeahan: Gotcha - You can feel free to migrate inline styles into the stylesheets so you can control per-layout, but need to add it to all the existing stylesheets if so. Should we just revert all these changes right here for now?

chat.php
```diff
@@ -197,13 +197,13 @@ function chat() {
...
+							<p style='display:inline-block; margin: 1px 0 3px;'>
```
levimeahan, line 206:
```suggestion
						<div style='display:block;flex-grow:1;'>
							<a style='display:inline-block;' href='{$system->links['members']}&user={$post['user_name']}' class='$class $statusType'>{$post['user_name']}</a><br />
							<p style='display:block; margin: 1px 0 3px;'>
```

chat.php
levimeahan:
```suggestion
					<div id='user_data_container' style='display: flex;flex-direction:row'>
```

chat.php
```diff
@@ -197,13 +197,13 @@ function chat() {
...
-					<div id='user_data_container' style='display: flex;flex-direction:row'>
+					<div id='user_data_container'>
 					    <div style='flex-shrink:0;'>
 					        <img style='max-height:45px;max-width:45px;' src='{$userData['avatar_link']}' />
                         </div>
 						<div style='display:block;flex-grow:1;'>
-							<a href='{$system->links['members']}&user={$post['user_name']}' class='$class $statusType'>{$post['user_name']}</a><br />
-							<p style='margin: 1px 0 3px;'>
+							<a style='display:inline-block;' href='{$system->links['members']}&user={$post['user_name']}' class='$class $statusType'>{$post['user_name']}</a><br />
+							<p style='display:block; margin: 1px 0 3px;'>
```
levimeahan, line 206:
```suggestion
							<p style='margin: 1px 0 3px;'>
```

levimeahan (approved):

PR #34, opened 2021-09-15
Avy in its own Column

MiteyDrill:
- Added avy column to the Chat page
 - Avy is now bigger
 
 
image

### DIFF
--- a/chat.php
+++ b/chat.php
@@ -161,12 +161,13 @@ function chat() {
 	// Table with chat posts
 	echo "<table class='table' style='width:98%;'>
 		<tr>
-			<th style='width:28%;'>Users</th>
-			<th style='width:61%;'>Message</th>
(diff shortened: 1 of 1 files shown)


levimeahan:
Hmm, I think this takes up too much horizontal space. We're already squeezing both chat messages and people with names a little on the long side

MiteyDrill:
Ok, would you like the avy column to be shorten a bit and the User column to take up the extra space so it has more breathing room, or do we leave things as they are atm

levimeahan:
> 
> 
> Ok, would you like the avy column to be shorten a bit and the User column to take up the extra space so it has more breathing room, or do we leave things as they are atm

I'd say leave things as they are atm, although we could try increasing avatar size a bit more. Maybe 50px

MiteyDrill:
Ok sounds good👍

PR #44, opened 2021-09-17
If Subject empty || less than 2 -> 'No Subject' in PM Subject Column

MiteyDrill:
-modified 1 line in try catch function

 *not sure if removing an exception line is the right way to do it, but it works, I think...

image
_sent with empty subject line_

### DIFF
--- a/privateMessages.php
+++ b/privateMessages.php
@@ -55,7 +55,8 @@ class Messaging {
 		try {
 			// Check minimum length 
 			if(strlen($subject) < self::MIN_SUBJECT_LENGTH) {
-				throw new Exception("Please enter a subject!");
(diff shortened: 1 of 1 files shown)


privateMessages.php
```diff
@@ -55,7 +55,8 @@ function validateForm() {
 		try {
 			// Check minimum length 
 			if(strlen($subject) < self::MIN_SUBJECT_LENGTH) {
-				throw new Exception("Please enter a subject!");
+				// throw new Exception("Please enter a subject!");
+				$subject = "No Subject";
```
levimeahan, line 59:
```suggestion
				$subject = "[No Subject]";
```

levimeahan (approved):

PR #58, opened 2021-09-19
Added support for numpad keys in battle hotkeys

NetzachSephira:
Exactly what the branch name and commit say.
Using the non-numpad keys was actually slower for me. Figured I'd add support for the numpad keys too.


MiteyDrill (commented):
Looks good! I'm not sure but I think you can also set the switch up like this->
case (48 || 96): key = 0;
  break;
case (49 || 97): key = 1;
  break;
.
.
.
etc but i'm not too sure I forgot most things about switch case grammar

classes/Battle.php
```diff
@@ -1487,24 +1487,44 @@ public function renderActionPrompt(Fighter $player, $default_attacks) {
                 switch (event.which){
                     case 48: key = 0;
                     break;
+                    case 96: key = 0;
+                    break;
```
levimeahan, line 1491:
```suggestion
                    case 48: 
                    case 96: 
                      key = 0;
                      break;
```

levimeahan: Let's use fallthrough cases like this to make it more concise

levimeahan:
> Looks good! I'm not sure but I think you can also set the switch up like this->
> case (48 || 96): key = 0;
> break;

Not quite, in this case `(48 || 96)` will evaluate to 48 and it becomes effectively just `case 48` . However fallthrough cases (case with no `break;`) will do what we want here

```javascript
switch(inputKeyCode) {
  case 48:
  case 96:
    key = 0;
    break;
}
```

MiteyDrill:
Okay cool I didn't know tht😎😎 ill make a mental footnote next time I'm
using a switch case

NetzachSephira:
Updated to use fallthrough cases

levimeahan (approved):

PR #83, opened 2021-09-25
Blue Scroll Layout

AlvinValens:
New layout "Blue Scroll" and all necessary files.


levimeahan (approved):
Sweet, LGTM! Will test it on my local and then push it

PR #85, opened 2021-09-26
Added the JBBcode library

MiteyDrill:
- http://jbbcode.com/
 - Can add custom [bb]code so in the future someone can add something for 
youtube links

 Unsure if $system->html_parse() is used for anything other than journals if that's the case there might be an issue somewhere but as far as I can tell the bb code seems to work in journals.
 
 Some tags are missing too like [em], I think. But it seems fairly easy adding a new [bb]code through the DefaultCodeDefinitionSet file

### DIFF
--- a/classes/System.php
+++ b/classes/System.php
@@ -431,64 +431,76 @@ class System {
     }
 
     public function html_parse($text, $img = false, $faces = false) {
-        $search_array = array(
-            "[b]","[/b]","[u]","[/u]","[i]","[/i]",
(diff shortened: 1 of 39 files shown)


MiteyDrill:
Hold up, just realized the pretty links from Chat come from html_parse and I commented that section. Have to fix that before pull can occur

MiteyDrill:
Too many issues implimenting bbcode library as is gonna have to do something else

PR #87, opened 2021-09-26
Fixed Blue Scroll Notifications

AlvinValens:
- Switch to Shadow Ribbon-style notifications.
- Shortened the banner.

### DIFF
--- a/index.php
+++ b/index.php
@@ -251,7 +251,7 @@ if($LOGGED_IN) {
 	}
 	if(!$player->global_message_viewed && !$ajax) {
 		$result = $system->query("SELECT `global_message`, `time` FROM `system_storage` LIMIT 1");
-		$results = $system->db_fetch($result, 'assoc');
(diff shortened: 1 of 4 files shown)


index.php
```diff
@@ -251,9 +250,11 @@
 		$player->global_message_viewed = 1;
 	}
 	if(!$player->global_message_viewed && !$ajax) {
-		$result = $system->query("SELECT `global_message` FROM `system_storage` LIMIT 1");
-		$message = $system->db_fetch($result)['global_message'];
+		$result = $system->query("SELECT `global_message`, `time` FROM `system_storage` LIMIT 1");
+		$results = $system->db_fetch($result, 'assoc');
```
levimeahan, line 254:
```suggestion
		$results = $system->db_fetch($result);
```
`'assoc'` is already the default so we don't need to specify it

levimeahan (approved):

PR #92, opened 2021-09-28
Settings Page Fixes

HitoriWoB:
Fixed error when changing password.
Fixed endless reload when changing layout settings.


settings.php
```diff
@@ -231,8 +231,13 @@ function userSettings() {
...
 			$system->query($query);
-			$system->message("Layout updated!<script type='text/javascript'>setTimeout('window.location.reload();', 2000);</script>");
-			$system->printMessage();
+			$system->message("Layout updated!");
+            $system->printMessage();
+            
+            // Only reload page if a layout change was made
+            if($layout != $player->layout) {
+                header("Refresh:2; url=$self_link");
+            }
```
levimeahan, line 240: You can't call `header()` when there's already been display (e.g. `$system->printMessage()` above). I think the answer here is to change from `window.location.reload()` to `window.location.assign(window.location.href)` as reload is probably resubmitting the form values too

secure/vars.sample.php
```diff
@@ -1,17 +0,0 @@
...
-$ENVIRONMENT = 'dev';
-$web_url = 'http://localhost/';
-
-$SC_OPEN = true;
-$register_open = true;
-
-$host = 'localhost';
-$username = 'database_username';
-$password = 'database_password';
-$database = 'database_name';
```
levimeahan, line 17: Can you revert this file deletion?

HitoriWoB:
Buggered out my repo, so I submitted a new request with necessary changes.

PR #105, opened 2021-10-01
Manual - Added just a tiny bit of stuff

MiteyDrill:
Goal isn't to write the manual. I'm not a very good writer.

But I want to see if adding an initial commit will encourage other users to start adding some information of their own into the manual. I'm still not sure what stats do or how equipment/effects work. Adding more information to the manual would be useful for newbie players and even regular players if it is fleshed out more.

Idk what the theory is called but I think its like 'if something is wrong/incorrect online someone will try correcting it' so maybe if someone sees the manual page is missing lots of information they'll feel inclined to add something interesting.


MiteyDrill:
image
Manual Page looks like this

levimeahan (approved):

Issue #97, opened 2021-09-30
Team page features/Changes 


SatoGojo:
https://i.imgur.com/Rd9AWWo.jpg
Basically allowing a team "signature banner" as a 2:1 ratio, moving mission details down to the mission tab, putting mission tab above team controls

Adding a hyperlink to profile page and team village page for said team 
For reference: 
https://i.imgur.com/mzHgYJR.jpg <- View profile page
https://i.imgur.com/oiGPuFx.jpg <- Village HQ Clan/Team page
Those hyperlinks would lead to showing team details: https://i.imgur.com/YniTE0u.png

levimeahan:
I like this idea

MiteyDrill:
I also like these ideas

Issue #55, opened 2021-09-19
Code in Team Boosts


SatoGojo:
A way to incentivize pvping at current stage of the game, could be something like, <Costs 25 Team points to unlock [boost]>

Use team points to unlock boosts for your squad

Skill boosts, AI yen boost/stat gain boost (maybe an increased % of point drop) ect ect

PR #111, opened 2021-10-02
Added Daily Tasks & Team Updates

AlvinValens:
- Daily Tasks generated every 24 hrs
- These tasks are randomly generated from a pool of choices, can be Missions, PVP, or AI and it could require battle wins instead of just battles.
- Rewards are also auto-generated based on the task's difficulty. This might require tweaking.
- All layouts were adjusted to include the Daily Task display.


AlvinValens:
#97 #55

AlvinValens:
Team update just requires a standard int(11) column in `teams` names `boost_time`

db/migrations/daily_tasks.sql
```diff
@@ -0,0 +1,49 @@
+-- phpMyAdmin SQL Dump
```
levimeahan, line 1: Can you add the date/time to this file name in format `yyyymmdd_hhmm_daily_tasks.sql`? and add the line

`update system_storage set database_version='yyyymmdd_hhmm';`
 Going to make some migration tooling that relies on prefixed file names.
 
 Also, add a row to add the `boost_time` column to teams table - Aiming for migrations to be painless/largely automated for everyone working on this

AlvinValens:
This should be fixed now.

training.php
```diff
@@ -59,6 +59,11 @@ function training() {
 				$train_length = $stat_extended_train_length;
 				$train_gain = $stat_extended_train_gain - 20;
 			}
+			// TEAM BOOST TRAINING GAINS
+			if ($player->team['boost'] == 'Training') {
+                $percent_increase = ceil($train_gain * $player->team['boost_amount'] / 100);
+                $train_gain += $percent_increase;
```
levimeahan, line 65: Hmm this one's a bit tricky since we have whole-number stats. I think I'm in favor of sticking with the time-based approach here since it scales to any level of % boost without needing to be concerned with rounding

team.php
```diff
@@ -118,7 +118,83 @@ function team() {
 		}
 		$system->printMessage();
 	}
-	else if($_GET['join_mission'] && $player->team['mission_id']) {
+	else if (isset($_POST['set_boost'], $_POST['set_amount'])) {
+
+		$boost_valid = true;
+		$Boost = $system->clean($_POST['set_boost']);
```
levimeahan, line 124: Is the intent of the capitalized first letter here to denote a variable derived from input (and maybe thus that should not be modified)? If so I think maybe `$_boost` would be a good way to set this

levimeahan:
Thanks for taking on all this! Left a couple comments - Also, if you can split each issue's changes (at least, for a medium+ issue) into its own PR to keep the size down and allow granular approving that would be appreciated 🙏

training.php
AlvinValens: I was partially conflicted on this. I just think it's a bit simpler to not mess with the training timers since it's more inconvenient to time your training and such when it's not an irregular amount. Like dealing with 5 hours 24 minutes instead of 6 hours, or 4mins 15secs instead of 5 minutes. In my opinion, 5 minutes is already pretty short lol.

team.php
AlvinValens: Not really a deep reason. It was late at night so I just randomly did it lol

training.php
levimeahan: Yeah, the problem is currently with a 2.5% training boost and an 8 point gain, you get 8.2 which rounds to 8.0

levimeahan: Alternatively can do % chance to get 1 extra stat point (but this needs to be on training completion, so that people can rapidly restart trainings)

levimeahan:
> 
> 
> Thanks for taking on all this! Left a couple comments - Also on future PRs, if you can split each issue's changes (at least, for a medium+ issue) into its own PR to keep the size down and allow granular approving that would be appreciated 🙏

Updated this comment to reflect - No need to split this PR apart, just future ones

team.php
levimeahan: Haha gotcha

AlvinValens:
Gotcha. I changed the team boosts to just have a chance of triggering a 25% boost gain. Obviously we can tweak this if it's too much but it's currently set at (10%, 20%, 30% chances of triggering 25% boost). @levimeahan

levimeahan (approved):

PR #118, opened 2021-10-06
added gender changes in ancient market

Algharna:
Screenshot_3
Screenshot_1
Screenshot_2


premium.php
```diff
@@ -207,6 +208,62 @@ function premium() {
...
+				throw new Exception("Invalid gender! {$new_gender}");
+			}
+
+	 if(!isset($_POST['confirm'])) {
+		echo "
+			<table class='table'><tr><th>Confirm Gender Change</th></tr>
+				<tr>
+					<td style='text-align:center;'>Are you sure you want to change your gender to $new_gender?<br />
+					<br />
+					<b>(IMPORTANT: This is non-reversable once completed, if you want to return to your original gender you will have to pay another fee.)</b>
```
levimeahan, line 234:
```suggestion
```

members.php
```diff
@@ -200,22 +200,27 @@ function members() {
...
+
+
 			echo "<td style='width:50%;'>
 			<label style='width:6em;'>Level:</label> 	$viewUser->level<br />
 			<label style='width:6em;'>Exp:</label> 	$viewUser->exp<br />
 			<label style='width:6em;'>Rank:</label> 	" . $ranks[$viewUser->rank] . "<br />" .
-			"<br />
-			<label style='width:6em;'>Gender:</label> 	$viewUser->gender<br />
+			"<br />";
+			if($viewUser->gender == "None") {
```
levimeahan, line 216:
```suggestion
			if($viewUser->gender == User::GENDER_NONE) {

```
Let's add constants for all the gender options on the User class

premium.php
```diff
@@ -967,6 +1024,25 @@ function premium() {
 		</form>
 		</td></tr>";
 
+		$genders = array('Male', 'Female','Non-binary','None');
```
levimeahan, line 1026: Let's make this array a static var on the User class. For reference, see: https://github.com/levimeahan/shinobi-chronicles/blob/ce15896764593a63a48892b7f393d3c7a0408fd6/classes/User.php#L188-L196

But you can do the constants instead of plain strings in defining that array

profile.php
```diff
@@ -302,34 +302,41 @@ function userProfile() {
 		<label style='width:$label_width;'>Next level in:</label> " . $exp_remaining . " exp<br />
 		<div style='height:6px;width:200px;border-style:solid;border-width:1px;'>" .
 			"<div style='background-color:#FFD700;height:6px;width:" . $exp_width . "px;' /></div>" . "</div>" .
-		"<br />
-		<label style='width:$label_width;'>Gender:</label> $player->gender<br />
+		"<br />";
+		if($player->gender == "None") {
+			echo "<label style='width:$label_width;'>Gender:</label> $player->gender
+			(Won't display)
```
levimeahan, line 308: I think the "(Won't display)" part is better to put on the gender selection/change screens, we can just leave this as "None"

register.php
```diff
@@ -197,7 +197,7 @@
         }
 
         // Gender
-        if(!($gender == "Male" || $gender == "Female")) {
+        if(!($gender == "Male" || $gender == "Female" || $gender == "Non-binary" || $gender == "None")) {
```
levimeahan, line 200: Once you make the $genders array on `User`, you can do

`if(!in_array($gender, User::$genders))`

register.php
```diff
@@ -266,6 +268,12 @@
...
             break;
+        case 'Non-binary':
+            $non_binary = 1;
+            break;
+        case 'None':
+            $none = 1;
+            break;
         default:
             break;
     }
```
levimeahan, line 279: We can take out this whole switch block

register.php
```diff
@@ -298,6 +306,8 @@
 		<label for='gender'>Gender</label><br />
 			<input type='radio' name='gender' value='Male' " . ($male ? "checked='checked'" : "") . " /> Male<br />
 			<input type='radio' name='gender' value='Female' " . ($female ? "checked='checked'" : "") . " /> Female<br />
+            <input type='radio' name='gender' value='Non-binary' " . ($non_binary ? "checked='checked'" : "") . " /> Non-binary<br />
+            <input type='radio' name='gender' value='None' " . ($none ? "checked='checked'" : "") . " /> None<br />
```
levimeahan, line 310:
```suggestion
foreach(User::$genders as $gender_opt) {
    echo "<input type='radio' name='gender' value='$gender_opt' " . ($gender == $gender_opt ? "checked='checked'" : "") . " /> $gender_opt<br />";
}
```

members.php
Algharna: done

premium.php
Algharna: done

profile.php
Algharna: done

register.php
Algharna: done

Issue #182, opened 2021-10-19
View battles - Draw


unownram:
Draw is clickable and attempts to find a user named such when clicked on. Can it be...idk made into normal text or summat

PR #184, opened 2021-10-19
fixed Draw url

Algharna:
Should work fine now
image

### DIFF
--- a/templates/viewBattles.php
+++ b/templates/viewBattles.php
@@ -15,12 +15,10 @@
             <td><a href="<?= $system->links['members']?>&user=<?= $battle['player1'] ?>" style='text-decoration:none'><?= $battle['player1'] ?></a></td>
             <td><a href="<?= $system->links['members']?>&user=<?= $battle['player2'] ?>" style='text-decoration:none'><?= $battle['player2'] ?></a></td>
             <td>
-            <?php if($battle['winner'] == Battle::DRAW): ?>
-                Draw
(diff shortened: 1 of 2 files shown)


Algharna:
#182

templates/viewBattles.php
```diff
@@ -15,10 +15,12 @@
...
             <td><a href="<?= $system->links['members']?>&user=<?= $battle['player2'] ?>" style='text-decoration:none'><?= $battle['player2'] ?></a></td>
             <td>
-                <?php if($battle['winner']): ?>
+            <?php if($battle['winner'] == Battle::DRAW): ?>
+                Draw
+            <?php elseif($battle['winner']): ?>
                     <a href="<?= $system->links['members']?>&user=<?= $battle['winner'] ?>" style='text-decoration:none'><?= $battle['winner'] ?></a>
                 <?php else: ?>
-                    <a href="<?= $self_link ?>&battle_id=<?= $battle['id'] ?>">Watch</a>
+                    <a href="<?= $self_link ?>&battle_id=<?= $battle['id'] ?>" style='text-decoration:none'>Watch</a>
```
levimeahan, line 23:
```suggestion
                    <a href="<?= $self_link ?>&battle_id=<?= $battle['id'] ?>">Watch</a>
```

levimeahan: We're pretty consistent everywhere else with "user names don't have underlines, all other types of links do" so I think we should preserve that here

Issue #183, opened 2021-10-19
Prevent attack if opponent has 0HP


HitoriWoB:
Reported by Loke.

Will be partially fixed by sending users to village upon AI loss, but if warring allows attacks inside village, this issue will remain outstanding.

PR #185, opened 2021-10-19
transported to village if you die/draw to ai

Algharna:
#183

### DIFF
--- a/arena.php
+++ b/arena.php
@@ -218,10 +218,7 @@ function arenaFight(): bool {
 			</td></tr></table>";
 			$player->health = 5;
             $player->ai_losses++;
-			$player->location = $player->village_location;
-			$location = explode('.', $player->location);
-			$player->x = $location[0];
-			$player->y = $location[1];
+			$player->moveToVillage();
             $player->battle_id = 0;
             $player->last_pvp = time();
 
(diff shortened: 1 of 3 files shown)


arena.php
```diff
@@ -216,7 +216,12 @@ function arenaFight(): bool {
             echo "<table class='table'><tr><th>Battle Results</th></tr>
 			<tr><td>You have been defeated.
 			</td></tr></table>";
+			$player->health = 5;
             $player->ai_losses++;
+			$player->location = $player->village_location;
+			$location = explode('.', $player->location);
+			$player->x = $location[0];
+			$player->y = $location[1];
```
levimeahan, line 224: Let's make this an instance method on the User class like `$player->moveToVillage()`, since we commonly have to do it

Algharna: done and updated it in arena/battles

arena.php
```diff
@@ -231,9 +233,13 @@ function arenaFight(): bool {
             echo "<table class='table'><tr><th>Battle Results</th></tr>
 			<tr><td>The battle ended in a draw. You receive no reward.
 			</td></tr></table>";
+			$player->health = 5;
+			$player->moveToVillage();
             $player->battle_id = 0;
             $player->last_pvp = time();
-            
+            $player->battle_id = 0;
+            $player->last_pvp = time();
```
levimeahan, line 241:
```suggestion
```

PR #201, opened 2021-10-28
Prevents user from continuing a combat mission outside the initial location

### LINKED ISSUE #132: Wrong message when attempting to move into an inaccessible area.
"You cannot travel into another village!"

This appears when you attempt to move out of the map, or when you try and move immediately after a loss.

shadekun:
When attempting to move before entering combat for a mission, the mission ends as if a combat draw occurs and tells the user to go back to the village.

Halves the yen gain that they have accumulated so far when they return to the village.

Coincidentally also closes #132

### DIFF
--- a/travel.php
+++ b/travel.php
@@ -70,9 +70,18 @@ function travel() {
             $player->location = $location;
             $player->y = $target_y;
             $player->x = $target_x;
+
+            if($player->mission_id && $player->mission_stage['action_type'] == 'combat') {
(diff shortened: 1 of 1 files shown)


levimeahan (approved):
LGTM - Can we add a warning on the travel page (where the "Go to Mission Location" button usually is) that warns people that moving will cancel the mission?

shadekun:
Added a warning in an h3 tag

PR #203, opened 2021-11-05
Premium bug fix

HitoriWoB:
Fixes:
Chat name color bug on premium page
Allows user admins to select red again

Patches:
Infinite level 1 forbidden seal


chat.php
```diff
@@ -49,15 +49,53 @@ function chat() {
...
 			$supported_colors = array(
-				'blue' => -1,
+				'black' => 0,'blue' => -1,
 				'pink' => -2,
 				'gold' => -3,
 				'green' => -4,
 				'teal' => -5,
 				'purple' => -6,
 				'red' => -7,
 			);
```
levimeahan, line 59: While we're here, can we get rid of this string => number mapping and just store the color names in the DB?

chat.php
```diff
@@ -49,15 +49,53 @@ function chat() {
...
+                case 'red':
+                    if(!$player->isUserAdmin() || !$player->isHeadAdmin()) {
+                        $player->chat_color = 'black';
+                    }
+                    break;
+                default:
+                    break;
+            }
+
+			$user_color = (!$player->canChangeChatColor()) ? 0 : ($supported_colors[$player->chat_color]);
```
levimeahan, line 98: per above
```suggestion
			$user_color = (!$player->canChangeChatColor()) ? 0 : ($player->chat_color);
```

chat.php
```diff
@@ -49,15 +49,53 @@ function chat() {
...
+                    }
+                    break;
+                case 'red':
+                    if(!$player->isUserAdmin() || !$player->isHeadAdmin()) {
+                        $player->chat_color = 'black';
+                    }
+                    break;
+                default:
+                    break;
+            }
```
levimeahan, line 96: Hmm what if we centralize the name colors as defined in premium.php to `User` and then just do an `if(!in_array($player->chat_color, $player->chat_color_options))` check?

https://github.com/elementum-games/shinobi-chronicles/blob/9874185125d13e20f86730561728812046576a11/premium.php#L974-L1029

premium.php
```diff
@@ -528,53 +528,53 @@ function premium() {
 		}
 		$system->printMessage();
 	}
-	else if(isset($_POST['change_color']) && ($player->forbidden_seal || $player->premium_credits_purchased)) {
+	else if(isset($_POST['change_color']) && $player->canChangeChatColor()) {
 		$color = $system->clean($_POST['name_color']);
 		switch($color) {
```
levimeahan, line 533: same comment as above - if we move color options to an array defined in `User`, then we can do a single array check here

HitoriWoB:
Changes made

classes/User.php
```diff
@@ -1254,6 +1273,66 @@ public function getAvatarSize(): int {
...
+        }
+
+        return false;
+    }
+
+    /**
+     * @param false $forceAll // should only be used for debugging
+     * @return string[]
+     */
+    public function getNameColors($forceAll = false) {
```
levimeahan, line 1299: Head admin should be able to use all colors, is this debug code worth pushing to production?

PR #204, opened 2021-11-06
Support System Upgrade.

HitoriWoB:
Upgraded support system as discussed.

Support type statics not being used everywhere
Added functionality for staff to re-open supports
Corrected bug where comments wouldn't go through when closing or opening a support

### DIFF
--- a/classes/SupportManager.php
+++ b/classes/SupportManager.php
@@ -113,9 +113,7 @@ class SupportManager {
             ];
 
             if($this->user->isModerator()) {
-                $types[] = [
-                    self::$TYPE_MOD_REQUEST,
-                ];
+                $types[] = self::$TYPE_MOD_REQUEST;
             }
         }
         return $types;
--- a/classes/SupportManager.php
+++ b/classes/SupportManager.php
@@ -1,14 +1,19 @@
 <?php
 
-/**
- * Class SupportManager
- *
- * Note: Set admin to true to fetch/update supports without restriction. Leave it false and set user ID/key
- * to automatically restrict search/update of functions that care
- * to only supports matching that ID/key
- */
 class SupportManager {
-    public static $validationConstraints = [
+    public int $user_id;
+    public string $user_name;
+    public int $user_support_level;
+    public string $ip_address;
+    public bool $staff;
+
+    public array $requestTypeUserlevels;
+    public array $requestPremiumCosts;
+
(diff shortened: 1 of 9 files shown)


classes/SupportManager.php
```diff
@@ -96,115 +91,91 @@ public function __construct($system, $user_id = 0, $user_support_level = false,
...
+                self::$TYPE_BALANCE,
+                self::$TYPE_REPORT_STAFF,
+                self::$TYPE_APPEAL_BAN,
+                self::$TYPE_SUGGESTION,
+                self::$TYPE_MISC,
             ];
-            if($staffLevel == System::SC_MODERATOR) {
-                $types[] = 'Mod Request';
+
+            if($this->user->staff_level == System::SC_MODERATOR) {
```
levimeahan, line 115:
```suggestion
            if($this->user->isModerator()) {
```

classes/SupportManager.php
```diff
@@ -385,163 +262,158 @@ public function supportSearch($variables = [], $limit = false, $order = [], $off
...
-     */
-    public function updateSupport($support_id, $admin_id = 0, $admin_name = '', $staff_level=0) {
-        // Validate support exists and belongs to user
-        if($this->admin) {
-            $support_data = $this->fetchSupportByID($support_id);
-        }
-        else {
-            $support_data = $this->fetchSupportByID($support_id, $this->user_id);
-        }
+    public function updateSupport($support_id, $support_staff = false) {
```
levimeahan, line 398: Could we replace the `$support_staff` parameter with `$this->user->isSupportStaff()`?

HitoriWoB: Will change this to `$this->staff` as that dictates if the instance of SupportManager is a staff connection or a user connection. 

My thought process here is:
Oreo has a support request entered and is support staff. Replies on the user end, I'd think sending a PM that her support was updated could be confusing.

HitoriWoB:
Made changes as reviewed.

HitoriWoB:
Upgrade is now ready to go.

classes/SupportManager.php
levimeahan: That makes sense - We can go ahead and merge this, but we should update that variable to reflect that. Possible names:

`$this->in_staff_mode`
`$this->viewing_as_staff`

levimeahan (approved):

PR #216, opened 2022-02-02
fixed images path not loading on travel page and no default chat colo…

godswillumukoro:
…r for user creation


levimeahan (approved):
LGTM, thanks!

PR #222, opened 2022-07-14
Cato memes

AlexKJones:
I think I fixed the background on the catjam gif


levimeahan:
Looks like it has a ghosting effect where the previous frames linger in the view now, can you try a different frame interpolation mode?

PR #224, opened 2022-08-09
Bug Fix - Prevent Survival Mission Spam in Village

Kengetsu:
Updated logic for survival missions to account for PVP combat while carrying out the mission. If player wins the PVP fight they can continue with the mission, however if they lose the mission will be considered a retreat and standard logic will be processed. Additionally, code was added to check if the player is supposed to be in survival combat but is somehow in the village (result of a defeat or otherwise) this was mainly added to prevent users already spamming the mission since the new check won't retroactively clear mission ids.

### DIFF
--- a/pages/missions.php
+++ b/pages/missions.php
@@ -168,7 +168,14 @@ function runActiveMission() {
                         else if($player->mission_stage['stage_id'] > 2){
                             $player->mission_stage['stage_id'] -= 1;
                         }
-                        $mission_status = $mission->nextStage($player->mission_stage['stage_id']);
+                        if ($player->location == $player->village_location)
+                        {
+                            $mission_status = $mission->nextStage($player->mission_stage['stage_id'] = 4);
+                        }
+                        else
+                        {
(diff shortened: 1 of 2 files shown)


pages/missions.php
```diff
@@ -168,7 +168,14 @@ function runActiveMission() {
...
                         }
-                        $mission_status = $mission->nextStage($player->mission_stage['stage_id']);
+                        if ($player->location == $player->village_location)
+                        {
+                            $mission_status = $mission->nextStage($player->mission_stage['stage_id'] = 4);
+                        }
+                        else
+                        {
+                            $mission_status = $mission->nextStage($player->mission_stage['stage_id']);
+                        }
```
levimeahan, line 178:
```suggestion
                        if ($player->location == $player->village_location) {
                            $player->mission_stage['stage_id'] = 4;
                        }  
                        
                        $mission_status = $mission->nextStage($player->mission_stage['stage_id']);
```

I think this way might be a little cleaner to read

levimeahan (approved):

PR #226, opened 2022-08-24
Regen Display Fix

### LINKED ISSUE #217: [Bug] Regen applies twice
Health/Chakra/Stamina regen applies twice every minute, once "automatically" if you're just sitting on the profile page and again on page refresh.

Kengetsu:
#217 - This isn't actually a bug, the server side code for applying regen actually applies 2:1 for health and a 1:1 the other stats. The display refresh code doesn't account for this thus it appears that you are getting regen twice for health except it's just on page refresh it properly pulls the value from the player object.

### DIFF
--- a/pages/profile.php
+++ b/pages/profile.php
@@ -241,39 +241,19 @@ function userProfile() {
 				remainingtime = 60;
 
                 //Check each bar to see if regen will exceed max.
-                if(statusBars.health.current + (regen * 2) > statusBars.health.max)
-                {
-                    statusBars.health.current = statusBars.health.max;
-                    statusBars.health.next_regen = statusBars.health.max;
-                }
-                else
-                {
-                    statusBars.health.current += (regen * 2);
-                    statusBars.health.next_regen = statusBars.health.current + (regen * 2);
-                }
+                let healthRegen = regen * 2;
+                
+                statusBars.health.current = Math.min(statusBars.health.current + healthRegen, statusBars.health.max);
+                statusBars.health.next_regen = Math.min(statusBars.health.current + healthRegen, statusBars.health.max);
                 
                 //Check Chakra Bar
-                if(statusBars.chakra.current + regen > statusBars.chakra.max)
-                {
-                    statusBars.chakra.current = statusBars.chakra.max;
-                    statusBars.chakra.next_regen = statusBars.chakra.max; 
-                }
-                else
-                {
-                    statusBars.chakra.current += regen ;
-                    statusBars.chakra.next_regen = statusBars.chakra.current + regen;  
-                }
+                statusBars.chakra.current = Math.min(statusBars.chakra.current + regen, statusBars.chakra.max);
+                statusBars.chakra.next_regen = Math.min(statusBars.chakra.current + regen, statusBars.chakra.max);
+                
                 //Check Stamina Bar
-                if(statusBars.stamina.current + regen > statusBars.stamina.max)
-                {
-                    statusBars.stamina.current = statusBars.stamina.max;
-                    statusBars.stamina.next_regen = statusBars.stamina.max; 
-                }
-                else
-                {
-                    statusBars.stamina.current += regen ;
-                    statusBars.stamina.next_regen = statusBars.stamina.current + regen;  
-                }
+                statusBars.stamina.current = Math.min(statusBars.stamina.current + regen, statusBars.stamina.max);
+                statusBars.stamina.next_regen = Math.min(statusBars.stamina.current + regen, statusBars.stamina.max);
+                
                 
                 //Update Health Bar
                 $('#health').html((statusBars.health.current === statusBars.health.max)? statusBars.health.current.toFixed(2) + '/' + statusBars.health.max.toFixed(2) : statusBars.health.current.toFixed(2) + '/' + statusBars.health.max.toFixed(2) + '-> <b style=\'color: green\'>' + statusBars.health.next_regen.toFixed(2) + '</b>');
@@ -296,7 +276,11 @@ function userProfile() {
 		//can't figure out why? its like the Regen changes every other minute in intervals or it doubles
 		//can't find the error with my script
 		//can't seem to find out where the error is, need help.
-
+		
+		function updateBarValue(bar)
+		{
+            
+		}
 		</script>
 		";
 
--- a/pages/profile.php
+++ b/pages/profile.php
@@ -235,6 +235,7 @@ function userProfile() {
                     if(statusBars[bar][0] + (regen * 2) > statusBars[bar][1])
                     {
                         statusBars[bar][0] = statusBars[bar][1];
+                        statusBars[bar][2] = statusBars[bar][1];
                     }
                     else
                     {
@@ -245,6 +246,7 @@ function userProfile() {
                 else if (statusBars[bar][0] + regen > statusBars[bar][1])
                 {
                     statusBars[bar][0] = statusBars[bar][1];
+                    statusBars[bar][2] = statusBars[bar][1];
                 }
                 else 
                 {
--- a/pages/profile.php
+++ b/pages/profile.php
@@ -234,7 +234,7 @@ function userProfile() {
 		setInterval(() => 
 		{
             
-			$('#regentimer').text(remainingtime); //minus 1 to compensate for lag
+			document.getElementById('regentimer').innerHTML = remainingtime; //minus 1 to compensate for lag
 
 
 			if(remainingtime <= 0){
--- a/pages/profile.php
+++ b/pages/profile.php
@@ -212,55 +212,80 @@ function userProfile() {
 		<script>
 		var remainingtime = " . (59 - $time_since_last_regen) . ";
         var statusBars = {
-            health: [{$player->health}, {$player->max_health}, 0],         
-            chakra: [{$player->chakra}, {$player->max_chakra}, 0],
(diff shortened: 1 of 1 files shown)


pages/profile.php
```diff
@@ -207,53 +211,56 @@ function userProfile() {
...
+            chakra: [{$player->chakra}, {$player->max_chakra}, 0],
+            stamina: [{$player->stamina}, {$player->max_stamina}, 0],     
+        };
 
 		var regen = {$player->regen_rate} + {$player->regen_boost}; //no regen cut
 
 		setInterval(() => {
 
-			document.getElementById('regentimer').innerHTML = remainingtime; //minus 1 to compensate for lag
+			$('#regentimer').text(remainingtime); //minus 1 to compensate for lag
```
levimeahan, line 224:
```suggestion
			document.getElementById('regentimer').innerHTML = remainingtime; //minus 1 to compensate for lag
```

Let's keep this as is, I'd like to drop jQuery soon and use native HTML methods

pages/profile.php
```diff
@@ -207,53 +211,56 @@ function userProfile() {
...
+                else 
+                {
+                    statusBars[bar][0] += regen;
+                    statusBars[bar][2] =  statusBars[bar][0] + regen;
+                }
+                //console.log(bar, statusBars[bar]);
+                if(bar.startsWith('max')) continue;
+                $('#' + bar).html((statusBars[bar][0] == statusBars[bar][1])? statusBars[bar][0].toFixed(2) + '/' + statusBars[bar][1].toFixed(2) : statusBars[bar][0].toFixed(2) + '/' + statusBars[bar][1].toFixed(2) + '-> <b style=\'color: green\'>' + statusBars[bar][2].toFixed(2) + '</b>');
+			    $('#' + bar + 'bar').val(statusBars[bar][0]);  
+            };
```
levimeahan, line 260: I actually think hardcoding the value setting for each bar is clearer than making an object and iterating through it here - Reading through this loop it's a bit difficult to understand what's going on with sections like

```
statusBars[bar][0] = statusBars[bar][1];
statusBars[bar][2] = statusBars[bar][1];
```

Could this block be refactored to use hardcoded statements for each bar type, so it's easier to read the logic you're trying to do here? I think it'll actually end up being less code and easier to read and see what's going on

pages/profile.php
```diff
@@ -207,53 +211,56 @@ function userProfile() {
...
-		var health = {$player->health};
-		var max_health = {$player->max_health};
-
-		var chakra = {$player->chakra};
-		var max_chakra = {$player->max_chakra};
-
-		var stamina = {$player->stamina};
-		var max_stamina = {$player->max_stamina};
+        var statusBars = {
+            health: [{$player->health}, {$player->max_health}, 0],
```
levimeahan, line 215: Could we restructure the arrays here to objects so that we can name the keys and preserve the semantic meaning? (also, not totally sure what the third element is for. Reading below, is this for the post-regen amount?)
```suggestion
            health: {
              current: {$player->health}, 
              max: {$player->max_health}, 
              ???: 0
            },         
```

pages/profile.php
```diff
@@ -163,37 +163,41 @@ function userProfile() {
...
+        $chakra_after_regen = ($player->chakra + ($player->regen_rate + $player->regen_boost - $regen_cut) > $player->max_chakra)? $player->max_chakra : $player->chakra + ($player->regen_rate + $player->regen_boost - $regen_cut);
+        $stamina_after_regen = ($player->stamina + ($player->regen_rate + $player->regen_boost - $regen_cut) > $player->max_stamina)? $player->max_stamina : $player->stamina + ($player->regen_rate + $player->regen_boost - $regen_cut);
+
+        $total_regen_rate = $player->regen_rate + $player->regen_boost - $regen_cut;
 
         echo "<td style='width:50%;'>
-		<label style='width:6.7em;'>Health:</label>" .
-            "<span id='health'>" . sprintf("%.2f", $player->health) . '/' . sprintf("%.2f", $player->max_health) . "</span><br />" .
+		<label style='width:6.7em;' for='healthbar'>Health:</label>" .
+            "<span id='health'>" . sprintf("%.2f", $player->health) .'/' . sprintf("%.2f", $player->max_health) . ($player->health != $player->max_health ? " -> <b style='color: green;'>" . sprintf("%.2f",$health_after_regen) . "</b>" : "") . "</span><br />" .
```
levimeahan, line 179: Hmm, why do we need to render the health after regen instead of rendering the current health?

levimeahan (commented):
Thanks for looking at this! Left a couple questions and proposed readability improvements for this fix in the comments

pages/profile.php
Kengetsu: Sure, I can do that. I turned it into an object to make it easier to set all of the proper values since you are doing multiple actions against the same bit of data.

pages/profile.php
Kengetsu: Yep, it is for the amount after next regen. I'll move it into an object.

pages/profile.php
Kengetsu: So it's actually doing both. So you see current health after regen but also the amount you will be at after the next regen. This allows at a glance to see how the regen is applied since there is no indication that the health regen is actually doubled.

levimeahan: Hmm maybe the better solution here is to update the regen rate indicator to show the different regen rates for each resource type? Example from WoB: 
image

levimeahan: My concern with showing the "current health -> health after next regen" is it could also be construed as "past health -> current health". By showing regen rate instead it's really clear which is which

Kengetsu: I can definitely see that concern, personally I like being able to see what my health will be at after I wait some time since I can immediately know if i'll be ready for combat by then or not. I feel like it's pretty clear what your current hp since it's inside the bar while this value points to what it will become.
image

levimeahan: Hmm yeah I think you're right, with the color coding and positioning after the max health that is pretty clear 👍 Okay, I'm on board

pages/profile.php
```diff
@@ -207,53 +211,81 @@ function userProfile() {
...
+                    statusBars.stamina.next_regen = statusBars.stamina.max; 
+                }
+                else
+                {
+                    statusBars.stamina.current += regen ;
+                    statusBars.stamina.next_regen = statusBars.stamina.current + regen;  
+                }
+                
+                //Update Health Bar
+                $('#health').html((statusBars.health.current === statusBars.health.max)? statusBars.health.current.toFixed(2) + '/' + statusBars.health.max.toFixed(2) : statusBars.health.current.toFixed(2) + '/' + statusBars.health.max.toFixed(2) + '-> <b style=\'color: green\'>' + statusBars.health.next_regen.toFixed(2) + '</b>');
```
levimeahan, line 279: We ensure above that `.current` is never higher than `.max`, so I think we can remove this logic. Actually it looks like the logic above is doing the same thing regardless of whether it's maxed or not?

```suggestion
                $('#health').html(statusBars.health.current.toFixed(2) + '/' + statusBars.health.max.toFixed(2) + '-> <b style=\'color: green\'>' + statusBars.health.next_regen.toFixed(2) + '</b>');
```

Kengetsu: I think making that change will cause the next regen to always be displayed which we don't want. I put the check in place to prevent the regen from showing once the bar is capped. Otherwise it will display.

levimeahan: Hmm I see, in that case how about we localize the ternary to the next regen preview section?
```suggestion
                let currentAndMaxHealth = statusBars.health.current.toFixed(2) + '/' + statusBars.health.max.toFixed(2) ;
                $('#health').html(currentAndMaxHealth + 
                  (statusBars.health.current === statusBars.health.max
                    ? ""
                    : '-> <b style=\'color: green\'>' + statusBars.health.next_regen.toFixed(2) + '</b>')
                );
```

Kengetsu: Interestingly that doesn't seem to work. 
image

`//Update Health Bar
                const currentAndMaxHealth = statusBars.health.current.toFixed(2) + '/' + statusBars.health.max.toFixed(2);
                $('#health').html(currentAndMaxHealth + (statusBars.health.current === statusBars.health.max) ? '' : '-> <b style=\'color: green\'>' + statusBars.health.next_regen.toFixed(2) + '</b>');
                $('#healthbar').val(statusBars.health.current);
                
                //Update Chakra Bar
                 const currentAndMaxChakra = statusBars.chakra.current.toFixed(2) + '/' + statusBars.chakra.max.toFixed(2);
                $('#chakra').html(currentAndMaxChakra + (statusBars.chakra.current !== statusBars.chakra.max)? ('-> <b style=\'color: green\'>' + statusBars.chakra.next_regen.toFixed(2) + '</b>') : '');
                $('#chakrabar').val(statusBars.chakra.current);	
                
                //Update Stamina Bar
                const currentAndMaxStamina = statusBars.stamina.current.toFixed(2) + '/' + statusBars.stamina.max.toFixed(2);
                $('#stamina').html(currentAndMaxStamina + (statusBars.stamina.current !== statusBars.stamina.max)? ('-> <b style=\'color: green\'>' + statusBars.stamina.next_regen.toFixed(2) + '</b>') : '');
                $('#staminabar').val(statusBars.stamina.current);	`

Kengetsu: Ok got it working, seems calling the concat function works since it does default casting to string.

levimeahan (approved):

PR #227, opened 2022-09-01
Clan Manual and Auto Kick

Kengetsu:
Added a button for manually kicking players from a clan seat after half of the auto kick time 15 days (using new const MAX_CLAN_HOLDER_IDLE_TIME / 2 ) and built a auto cleanup function that can be called in the weekly_tasks file.


classes/System.php
```diff
@@ -660,7 +661,15 @@ public function time_remaining($timestamp): string {
...
-        $contents = $this->clean($contents);
+
+        if (is_array($contents))
+        {
+            $contents = json_encode($contents);
+        }
+        else
+        {
+            $contents = $this->clean($contents);
+        }
```
levimeahan, line 672: json_encode won't strip out single quotes that are used for SQL injection, so we still need to do that since we have no guarantees about the data passed into `log` (could include user input)

```suggestion
        if (is_array($contents))
        {
            $contents = json_encode($contents);
        }

       $contents = $this->clean($contents);
```

db/migrations/20220830_clan_stored_routine.sql
```diff
@@ -0,0 +1,51 @@
+START TRANSACTION;
```
levimeahan, line 1: Let's remove this file since we have the regular PHP version

pages/clan.php
```diff
@@ -116,6 +120,61 @@ function clan() {
 			$system->printMessage();
 		}
 	}
+    //Kick Controls
+    if($page == 'kick' && $_GET['kick']) {
```
levimeahan, line 124: TBH I think we can just do auto kick and leave out the manual kick option. If we think it's okay for users to remove someone after a period of time we should just proactively remove them

classes/System.php
Kengetsu: I definitely understand the concern there. I can just manually log the output of the clean clan function since I added that piece in there since I was using it to log the return but running clean on it makes it hard to read or search against if you needed to query the log. 
image

levimeahan (approved):
Thanks for addressing comments, looks good!

PR #231, opened 2022-11-26
[Issue-161] Multibuy for consumables

KeanuBlee:
Added an option to immediately purchase enough to get the user to the max amount of a given consumable. Let me know if there's any issue!

### DIFF
--- a/pages/store.php
+++ b/pages/store.php
@@ -62,9 +62,23 @@ function store() {
 				throw new Exception("You already own this item!");
 			}
 			
+			$max_missing = -1;
+			if (isset($_GET['max'])) {
+				if ($player->checkInventory($item_id, 'item')) {
+					$max_missing = $max_consumables - $player->items[$item_id]['quantity'];
+				} else {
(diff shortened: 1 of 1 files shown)


pages/store.php
```diff
@@ -62,10 +62,24 @@ function store() {
 				throw new Exception("You already own this item!");
 			}
 			
+			$max_missing = -1;
+			if (isset($_GET['max'])) {
```
levimeahan, line 66: Could we move all the new logic into this if block? I think it would be easier to maintain if we have all the logic for purchasing max items and all the logic for purchasing one item isolated in their own logic blocks. 

As is I'm concerned that it's not clear what values of `max_missing` can cause a user to buy multiple consumables, with the multiple logic blocks below interacting with it

levimeahan (commented):
Thanks for taking this on! The logic mostly makes sense but I would really like to see a reorganization into distinct max vs single purchase paths (see prev comment) before we merge this

levimeahan (approved):

PR #232, opened 2022-12-07
[WIP] Add docker compose for mysql setup.

tomaszbawor:
Initial look of how I would imagine dockerization for development purposes. Right now it is just the database but the goal is to run "docker compose up -d" to start the whole development environment (MySQL, PHP, composer) 

Looking for suggestions and if you like the idea


levimeahan (approved):
Looks good, thanks!

PR #233, opened 2022-12-11
Docker environment

tomaszbawor:


db/migrations/V1_4__20210926_2113.sql
```diff
@@ -12,7 +12,7 @@ UPDATE `ai_opponents` SET
     `strength` = 0.01,
     `endurance` = 0.01,
     `intelligence` = 0.1,
-    `willpower` = 0.05;
+    `willpower` = 0.05 where 0=0;
```
levimeahan, line 15: Huh this is odd, is this auto-generated?

levimeahan (approved):

db/migrations/V1_4__20210926_2113.sql
tomaszbawor: Sorry this is something that I have added in order to prevent DB errors when they block Updates without Where clause

PR #235, opened 2022-12-31
Profile Page - Radial Graph for User Stats

MiteyDrill:
Radial Graph on profile page showcasing players current range of proficiency among skills - range exists between 1-5
image
image


levimeahan:
Whoa, this is cool! Can we make it smaller though? My thought is we could drop the "Range of skill proficiency title" and make it smaller and put it here (in this screenshot I also changed the profile left/right columns from 50/50 to 45/55): 
image

MiteyDrill:
Yeah that's a good idea I'll try putting it there

MiteyDrill:
image
Position Updated

levimeahan (approved):

MiteyDrill:
oops lol

PR #236, opened 2023-01-02
Mist tribe

TylerRedd:


levimeahan:
Resize this to max 70px tall and I'll approve 👍

TylerRedd:
> Resize this to max 70px tall and I'll approve 👍

Done. But my option to delete the bigger one is greyed out.

PR #238, opened 2023-01-08
User Stats/Pools are aligned Center on Arena Page

MiteyDrill:
User Stats are aligned left on the arena page, I added a <div> container surrounding the pools and aligned them center using inline css - its not a huge change but visually it might make the arena page a little more organized.

image

I didn't do a lot of testing to see if it works for all ranks and scenarios. I tried the arena a few times and it seems to be working on my local, but I had to surgically move the code from another branch into this fresh one so if there's any errors it might be because of that.

### DIFF
--- a/templates/battle/battle_interface.php
+++ b/templates/battle/battle_interface.php
@@ -42,22 +42,47 @@
     }
 
     .resourceBarOuter {
-        height:6px;
-        width:250px;
+        position: relative;
+        height: 17px;
+        width: 250px;
         border-style:solid;
-        border-width:1px;
+        border-radius: 12px;
+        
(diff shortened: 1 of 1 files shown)


levimeahan:
Hmm this seems to create a larger gap on the left than on the right:
image

Any chance you'd actually be willing to move the numbers inside the bars, like this?
image

MiteyDrill:
I'll look into it

MiteyDrill:
image
Updates pool bars, I might add some extra css tomorrow to make them just a bit shinier and see if I can make the bar borders thinner.

MiteyDrill:
image
Updated CSS

levimeahan (approved):

PR #240, opened 2023-01-10
Fix Overlapping Messages

MiteyDrill:
- small patch in Chat.php to prevent message from showing up on refresh
- some css changes in how system messages is displayed on Geisha and Shadow Ribbon

### DIFF
--- a/pages/chat.php
+++ b/pages/chat.php
@@ -160,6 +160,8 @@ function chat() {
 			echo "<script type='text/javascript'>
 			var refreshID;
 			$(document).ready(function(){
+				const system_message =  document.querySelector('div');
+				div.classList.remove('socialPosts'); //quick fix - removes system messages from appearing twice on social chat page
 				refreshID = setInterval('javascript:$(\'#socialPosts\').load(\'$self_link&request_type=ajax\');', 3000);
 			});
 			</script>";
--- a/style/geisha/geisha.css
+++ b/style/geisha/geisha.css
@@ -409,12 +409,14 @@ p {
 }
 
 .systemMessage {
-	position: absolute;
-	margin: 0;
-	width: 670px;
-	
-	left: 0;
-	top: 24px;
+	/* position: absolute; */
+	margin: 10px 15px;
+	padding: 20px 0;
+
+	background-color: rgb(55, 47, 38);
+
+	border: solid rgba(20, 20, 20) 1px;
+	border-radius: 5px;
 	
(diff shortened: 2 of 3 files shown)


MiteyDrill:
image
issue example:

MiteyDrill:
Geisha:
image
image
Shadow Ribbon:
image
image

levimeahan:
I like the removal of overlapping messages, but one problem here is that now for all system messages, substantially more vertical space is taken for each one (and our layout is already quite stretched vertically, especially in combat).  Could we do this in a way that doesn't push the rest of the page contents down?

MiteyDrill:
Its possible to use a position absolute to set the systemMessage to any other part of the screen - or another solution would be using jquery or javascript to make the message bar act like a popup or one of those fancy notification pop ups that appear on the corner of the screen?

Currently this seems like the simplest solution though, and the vertical increase doesn't seem so dramatic that it should ruin user experience. I think also if users notice the vertical increase they'll be inclined to check the message bar to see why their screen was pushed down.

Otherwise if its okay w/ you it could be fun setting up one of those fancy notification pop ups but it might take me a while to do so

levimeahan:
> Currently this seems like the simplest solution though, and the vertical increase doesn't seem so dramatic that it should ruin user experience. I think also if users notice the vertical increase they'll be inclined to check the message bar to see why their screen was pushed down.

The other problem here is this messes with muscle memory, if users are used to clicking in a certain place for e.g. training or buying stuff, it will get pushed out of place. 

What if we kept the box with background the way you've created it, but absolutely positioned it in the green area so it doesn't interrupt page content, and if necessary let it flow upwards into the yellow region?
image

MiteyDrill:
Yeah i'll look at it tomorrow and try to make these changes

MiteyDrill:
image
image
image
Here's an example of the Geisha CSS Code

I found a way to set it using absolute but the notification bar is pretty static and will have to be updated in each css file to fit the specific system message area but it shouldn't take too long to do that just a few minutes of tinkering.

MiteyDrill:
image
The system message bar does look a little funny like here, but while checking in most places I think it'll be okay for regular users.

HitoriWoB:
Legacy layouts are being boxed up soon.
A noble salute to you, Overlapping Messages!
Though you may be gone.... you will never be forgotten.

MiteyDrill:
o7

PR #245, opened 2023-01-15
Survival missions now lock the user in battle

Kengetsu:
This will lock players in battle during survival missions preventing them from being attacked or accidentally moving without retreating.

### DIFF
--- a/classes/Mission.php
+++ b/classes/Mission.php
@@ -108,6 +108,7 @@ class Mission {
             if($this->mission_type == 5) {
                 $this->current_stage['ai_defeated'] = $this->player->mission_stage['ai_defeated'] ?? 0;
                 $this->current_stage['mission_money'] = $this->player->mission_stage['mission_money'] ?? 0;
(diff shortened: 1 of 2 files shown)


Kengetsu:
Still working on an inconvenience with the enter combat button. Once clicked you need to manually refresh the combat page for the next opponent to show.

levimeahan:
Hmm before we merge this, I'm not sure it's desirable to allow users to spam survival mission repeatedly with no threat of being attacked. I'd be okay with blocking movement until the mission is done if you want to change it to that, but otherwise I need to give some design thought to survival missions

Kengetsu:
That's actually what this is doing. It will lock them in combat with the previous AI. So they can't keep spamming AI and the mission doesn't progress until they press enter battle or retreat. It's meant as a way to prevent the user from doing anything else while they are "ambushed"

levimeahan (approved):
Oh I see, so they can't heal between fights. Okay, that makes more sense! Let's do it

PR #249, opened 2023-01-16
Profile page UI pools

MiteyDrill:
Slightly updated the pool bars on the profile page
image
Ran into an issue when trying to work on the preview regen pool gain bar - the pool bars are made with the <progress> html component which is a little funky and won't let me edit in or add a preview bar without a lot of extra work - so if its okay I might just replace the <progress> html with a simple <div> and then go in and add in the preview bar tomorrow or later this week. I think it won't be too complex to do the switch but I won't be able to tell until tomorrow i'm going to sleep lol


MiteyDrill:
oh also in my commit a little syntax error popped thru, should be fixed in my next commit I tried to discard that change but it wouldn't revert for some odd reason
image

levimeahan:
Hmm, how are these different from the current display? Trying to understand the comparison before we merge this

MiteyDrill:
image
Currently the health bars have a static set width and they also appear to be floating left of their container with a lot of open space to the right of them. I don't remember everything I did but I think I just made 2 or 3 small changes.

I surrounded the bars in their own respective containers and set their width to 100% to make them more responsive to any parent container changes.

I also went in and cleaned up some of the text formatting and put floating text with missing < > containers within their own respective containers like <span>

image

Also the stats now float to the right of their containers with - justify-content: space between 
image
image
I added some margin to the right of the stat text but it's been a few days since making it and I kinda am not fond of the margin. I don't remember why I added that little space

levimeahan (approved):
sweet, let's ship it

PR #257, opened 2023-01-27
quick spacing fix

AlexKJones:
Noticed that the new hat memes had an extra tab preventing use


levimeahan:
Ahh the whitespace here actually wasn't an issue, I just hadn't deployed the latest code to the production server yet. Should be good now on live!

PR #258, opened 2023-01-28
fix: updated manual UI and added bloodline list

jayantamadhav:
Made the manual prettier
Included bloodline_list.php in Bloodline section, because as a Academy student I had no way of getting any BL information and it should be included in manual.


levimeahan (approved):
Thanks for updating!

jayantamadhav:
I just realized the bloodlines listing is not showing up, will send a fix in next week

PR #260, opened 2023-01-30
Staff rework

HitoriWoB:
Mod panel upgrade is complete. Ready for review.

### DIFF
--- a/classes/ReportManager.php
+++ b/classes/ReportManager.php
@@ -1,0 +1,247 @@
+<?php
+
+class ReportManager {
+    public User $player;
+    public System $system;
+    public bool $staff_connection;
+
+    const REPORT_TYPE_PROFILE = 1;
+    const REPORT_TYPE_PM = 2;
+    const REPORT_TYPE_CHAT = 3;
+
+    const VERDICT_UNHANDLED = 0;
+    const VERDICT_GUILTY = 1;
+    const VERDICT_NOT_GUILTY = 2;
+
+    const MAX_NOTE_SIZE = 1000;
+
+    public static array $report_verdicts = array (
+        self::VERDICT_UNHANDLED => 'Unhandled',
+        self::VERDICT_GUILTY => 'Guilty',
+        self::VERDICT_NOT_GUILTY => 'Not Guilty'
+    );
+
+    public static array $report_types = array(
+        self::REPORT_TYPE_PROFILE => 'Profile/Journal',
+        self::REPORT_TYPE_PM => 'Private Message',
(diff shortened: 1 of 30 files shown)


classes/StaffManager.php
```diff
@@ -0,0 +1,575 @@
...
+    const STAFF_LOG_SUPPORT = 'support_action';
+
+    public static array $verdicts = [
+        self::VERDICT_UNHANDLED => 'Unhandled',
+        self::VERDICT_GUILTY => 'Guilty',
+        self::VERDICT_NOT_GUILTY => 'Not Guilty'
+    ];
+
+    public function __construct(User $player) {
+        $this->player = $player;
```
levimeahan, line 83: In the interest of avoiding a recursive data structure, since this `player` property doesn't seem to actually be used anywhere else in this class, maybe we just store the individual variables and not the instance of the whole class?

classes/StaffManager.php
```diff
@@ -0,0 +1,575 @@
...
+        return self::$staff_level_names[$staff_level][$name_type];
+    }
+
+    public function getBanLengths() {
+        $ban_lengths = [
+            1 => '1 Day',
+            7 => '1 Week',
+            30 => '1 Month',
+            90 => '3 Months'
+        ];
```
levimeahan, line 496: Ideally let's move this to a const or static variable at the top of the class so it's easier to find and edit

classes/User.php
```diff
@@ -92,6 +99,7 @@ class User extends Fighter {
 
     public $exp;
     public $staff_level;
+    public bool|StaffManager $staff_manager;
```
levimeahan, line 102: Generally if a class is either present or not, I think nullable can be a nicer way to represent that, you can do `?StaffManager`

levimeahan (approved):
yolo

classes/ReportManager.php
```diff
@@ -0,0 +1,247 @@
...
+    public static array $report_types = array(
+        self::REPORT_TYPE_PROFILE => 'Profile/Journal',
+        self::REPORT_TYPE_PM => 'Private Message',
+        self::REPORT_TYPE_CHAT => 'Chat Post'
+    );
+    public static array $report_reasons = array(
+        'Spamming', 'Harassment', 'Explicit Language/Content'
+    );
+
+    public function __construct($system, $player, $staff_connection = false) {
```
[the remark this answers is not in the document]
HitoriWoB: Used as a control member that only allows certain methods to be used if the manager is loaded for a staff member.

PR #262, opened 2023-02-06
Even more memes

TylerRedd:
Did a few more requests, a couple missing one's and a stitch gif I made. I'm done for a while.


levimeahan (approved):
Thanks for adding! Now that https://github.com/elementum-games/shinobi-chronicles/pull/265 has merged you don't even need to add codes to memes.php anymore, it will use filename as the code

PR #264, opened 2023-02-11
Premium updates

HitoriWoB:
Updated premium page to use templates.

Changed Ancient Market to allow users to not only place their AK in an offer, but users can now place their Yen in an offer. A player's yen offer can be accepted by another player wishing to sell their AK without them having to place the AK on the market.

Added the display that shows the yen cost per AK on the market (regardless of offer type).


HitoriWoB:
This update also gives four dragons +5 stats/min for stat transfers.

Chunin gain 5 more
Jonin gain 5 more

This stacks to bring the total up to 25 stats/min as a maximum at Jonin.

pages/premium.php
```diff
@@ -173,102 +181,81 @@ function premium() {
...
+                Doing this will also change your login name to the name you select.<br />
+                Changing your name will make it available for use to anyone!";
+                $additional_form_data = ['new_name' => ['input_type'=>'text', 'value'=>$new_name]];
+                $submit_value = "name_change";
+                $button_value = "Confirm Change";
 
-			$sql = "UPDATE `users` SET `user_name` = '%s',  `premium_credits` = `premium_credits` - %d, `username_changes` = `username_changes` - %d WHERE `user_id` = %d LIMIT 1;";
-			$system->query(sprintf($sql, $new_name, $akCost, $nameCost, $player->user_id));
-			$player->subtractPremiumCredits($akCost, "Username change");
+                require 'templates/premium/purchase_confirmation.php';
```
levimeahan, line 210: We spend so much effort here declaring variables for the template, I wonder if there's a better way...

templates/premium/character_reset_complete.php
```diff
@@ -0,0 +1,12 @@
+<?php
+/**
+ * @var string $self_link
+ */
+?>
+
+<table class='table'><tr><th>Character Reset</th></tr>
+    <tr><td style='text-align:center;'>
+        You have reset your character.<br />
+        <a href='<?=$system->link?>?id=1'>Continue</a>
```
levimeahan, line 10:
```suggestion
        <a href='<?= $system->links['profile'] ?>'>Continue</a>
```
I think ideally we should start moving to this pattern at least for common links. This gives us flexibility

levimeahan (approved):

PR #265, opened 2023-02-14
Update getMemes function to scan directory

Kengetsu:
Replace getMemes with new function to automatically scan the dir and create the relevant text replacement codes. Didn't bother with creating a public/private directory so I just flattened out the images. All codes will be based on the img files in the directory with - and _ removed and forced lowercase.


levimeahan (approved):
Looks good, thanks!

PR #269, opened 2023-02-26
Cosmetic & grammatical fixes

HitoriWoB:
Fixed cosmetic display issues for new staff menus.
Fixed error in not removing elements when necessary for stat cuts Fixed Inbox to allow special char display
Fixed table display issue in user records
Fixed grammatical errors in staff logs
Added globals to staff logs

### DIFF
--- a/classes/Inbox.php
+++ b/classes/Inbox.php
@@ -558,6 +558,7 @@ class Inbox {
         $convos_data = $system->db_fetch_all($result);
 
         foreach($convos_data as $convo) {
+            $convo['title'] = htmlspecialchars_decode($convo['title'], ENT_QUOTES);
 
             // Get a list of members for each conversation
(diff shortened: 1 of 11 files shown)


classes/Inbox.php
```diff
@@ -558,6 +558,7 @@ public static function allConvosForUser($system, $user_id): array {
         $convos_data = $system->db_fetch_all($result);
 
         foreach($convos_data as $convo) {
+            $convo['title'] = htmlspecialchars_decode($convo['title'], ENT_QUOTES);
```
levimeahan, line 561: Because the Inbox page where this gets displayed is React-based this is okay, but we should be extremely careful with this as a general pattern. 

Any desanitization of data that gets displayed in a PHP template or in React UI using `dangerouslySetInnerHtml` can introduce XSS vulnerabilities

levimeahan (approved):

PR #273, opened 2023-03-03
Fixes for avatar size and long name wrapping.

Kengetsu:
Fix some overflowing of div in chat when images are too large. Also fix css getting ignored on battle page for avatar sizes.

image

### DIFF
--- a/templates/chat.php
+++ b/templates/chat.php
@@ -16,15 +16,20 @@
             flex-direction: row;
         }
         #user_data_container .avatarContainer {
-            max-width: 75px;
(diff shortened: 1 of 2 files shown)


templates/chat.php
```diff
@@ -16,15 +16,20 @@
             flex-direction: row;
         }
         #user_data_container .avatarContainer {
-            max-width: 75px;
+            max-width: clamp(45px, 100%, 75px);
```
levimeahan, line 19: Whoa, this is cool! TIL

levimeahan (approved):
Thanks for the fix!

Issue #142, opened 2021-10-11
Travel updates (map size / UI)


levimeahan:
- Increase map size
- CD on traveling?
- Travel page UI revamp
  - Add auto-refresh 
  - User targeting system?

Issue #141, opened 2021-10-11
Add spar arena and regen font map squares ($50)


levimeahan:
**Spar Arena**
Square on the map where no PvP except spar can take place.
- Block arena fights too

**Regen Font**
- Change village regen boost from +100% => +50%
- Make a square on the map where regen is increased by +100%

levimeahan:
Bounty: $50

Issue #107, opened 2021-10-02
Add auto refresh in travel page ($300)


Rhaegarrr:
Add auto refresh in combat and travel page

levimeahan:
Bounty: $300

**Acceptance Criteria**
TODO: Add example structure for dynamic interfaces, to clarify how to approach this
- Interface ported to ReactJS
- Back-end API endpoint that returns JSON created for moving and viewing nearby users
- Traveling should use the endpoint
- The nearby player list should automatically refresh using the endpoint to fetch the latest list of available users, every 500ms

PR #274, opened 2023-03-06
Travel-Overhaul

AlvinValens:
React-based Travel and Scout area display that update without a page reload.

**Travel**
- Map are generated from the database and can be set with dynamic tile size, pixel size.
- Locations are pulled from the database and are linked to the map.
- Locations support both background colors and images.
- Locations can lock out PVP or AI fights.
- Locations can have increased regeneration rates.
- "Portals" can be created to link a map to another. Can be done dynamically or by admin in the future.
- Keyboard shortcut support for arrows and w, a, s, d.
- Movement delays for Death, PVP or AI fight, or just moving.

**Scout Area**
- Ability to filter our specific ranks.
- Refreshes every 500ms and displays players nearby depending on both scout range and stealth.
- Attack links are auto-generating every X minutes.

**Misc**
- X.Y was changed to X.Y.Z to support multiple maps. Currently everything would remain the same.
- All menu options are now shown regardless of location and are hidden by CSS.
- Added Inbox styling to Rainbow layout.

https://i.imgur.com/Zx4B9kq.png
https://i.imgur.com/Jnujxz6.png

#107 #141 #142


classes/TravelManager.php
```diff
@@ -0,0 +1,261 @@
...
+        foreach ($users as $user) {
+            // check if the user is nearby (including stealth
+            $scout_range = max(0, $this->user->scout_range - $user['stealth']);
+            $location = new TravelCoords($user['location']);
+            if ($location->z !== $this->user->location->z ||
+                abs($location->x - $this->user->location->x) > $scout_range ||
+                abs($location->y - $this->user->location->y) > $scout_range) {
+                continue;
+            }
+            $user['location_array'] = [$location->x, $location->y, $location->z];
```
levimeahan, line 233: Generally a fan of keeping this structured so we don't get x/y mixed up as y/x by mistake, could we make this explicitly `$user['x']`, `$user['y']`, `$user['map_id']`?

classes/User.php
```diff
@@ -587,7 +611,7 @@ public function loadData($UPDATE = User::UPDATE_FULL, $remote_view = false): str
         if($this->system->db_last_num_rows != 0) {
             $result = $this->system->db_fetch($result);
             $this->village_location = $result['location'];
-            if($this->location === $this->village_location) {
+            if($this->location->fetchString() === $this->village_location) {
```
levimeahan, line 614: similar comment as previous - we should move to a comparison format that isn't vulnerable to typecasting errors, like `x:y:z` or a dedicated `equals()` method

classes/Travel.php
```diff
@@ -0,0 +1,98 @@
...
+
+    public static function checkPVPDelay(int $last_pvp): int {
+        $diff = time() - $last_pvp;
+        return self::TRAVEL_DELAY_PVP - $diff;
+    }
+
+    public static function checkAIDelay(int $last_ai): int {
+        $diff = time() - $last_ai;
+        return self::TRAVEL_DELAY_AI - $diff;
+    }
```
levimeahan, line 62: I think these two need to use milliseconds like the movement delay does, so we have full control over how long this window is.

templates/profile.php
```diff
@@ -360,7 +360,7 @@ function round(num, places) {
     <br />
     <label style='width:<?= $label_width ?>;'>Gender:</label> <?= $player->gender ?><br />
     <label style='width:<?= $label_width ?>;'>Village:</label> <?= $player->village ?><br />
-    <label style='width:<?= $label_width ?>;'>Location:</label> <?= $player->location ?><br />
+    <label style='width:<?= $label_width ?>;'>Location:</label> <?= $player->location->fetchString() ?><br />
```
levimeahan, line 363: Seems we'd want to give players a more descriptive representation of what map they're on. Maybe we do just X.Y for now and later add in map name? Or map name if it's not the default

ui_components/src/travel/ScoutArea.js
```diff
@@ -0,0 +1,102 @@
...
+    player_data,
+    membersLink,
+    attackLink,
+    view_as,
+    view_genin,
+    view_chuunin,
+    view_jonin
+}) => {
+
+    if (player_data.name === 'Akademi-sei' && view_as === false) {
```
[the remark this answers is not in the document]
levimeahan: Also, ideally we would use Rank ID as the comparison here so if we were to tweak rank names a little bit this wouldn't break

levimeahan (approved):
Wow, heavy lift and the could largely looks good! 

**Blocking Feedback**
Really the only feedback I have here that's blocking is location comparisons. I'd really like to either shift from `x.y.z` to `x:y:z` format so there's no chance it gets typecast to an int and 1.1 == 1.10, or implement a dedicated `equals()` method. Given the way the strings are used as location keys, I think changing the string format is probably easier

classes/Mission.php
```diff
@@ -129,7 +129,7 @@ public function nextStage($stage_id) {
         if($this->current_stage['action_type'] == 'travel' || $this->current_stage['action_type'] == 'search') {
             for($i = 0; $i < 3; $i++) {
                 $location = $this->rollLocation($this->player->village_location);
-                if(!isset($villages[$location]) || $location == $this->player->village_location) {
+                if(!isset($villages[$location->fetchString()]) || $location->equals(TravelCoords::fromDbString($this->player->village_location))) {
```
levimeahan, line 132: If we made `player->village_location` also an instance of `TravelCoords` we could save a lot of effort on all the places that check it, and further enhance our type safety

classes/TravelCoords.php
```diff
@@ -0,0 +1,35 @@
...
+
+    public function fetchString(): string {
+        return $this->x . ':' . $this->y . ':' . $this->map_id;
+    }
+
+    public function distanceDifference(TravelCoords $target_coords): int {
+        $diff_x = abs($this->x - $target_coords->x);
+        $diff_y = abs($this->y - $target_coords->y);
+        return max($diff_x, $diff_y);
+    }
```
levimeahan, line 24: 🎉

classes/User.php
```diff
@@ -481,10 +485,31 @@ public function loadData($UPDATE = User::UPDATE_FULL, $remote_view = false): str
             array_unshift($this->stats, 'bloodline_skill');
         }
 
-        $this->location = $user_data['location']; // x.y
-        $location = explode(".", $this->location);
-        $this->x = $location[0];
-        $this->y = $location[1];
+        if ($user_data['location'])
```
levimeahan, line 488: Where should the `{ }` be for this if statement?

db/migrations/20230306004011_travel_overhaul.php
```diff
@@ -0,0 +1,208 @@
...
+                    [$new_loc, $user_data['user_id']]
+                );
+            }
+        }
+
+        // add map_id to user location
+        $this->execute("
+            UPDATE `users`
+                SET `location` = concat(`location`, ':1');
+        ");
```
levimeahan, line 126: The PHP way there totally works, but FYI there is a much simpler SQL alternative: https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_replace
```suggestion
  $this->execute("UPDATE `users` SET `location`=
    CONCAT(
      REPLACE(`location`, '.', ':'), 
      ':1'
    )
  ");
```

levimeahan: And for `down()` you can use 
```mysql
UPDATE users SET location = REPLACE(
  SUBSTRING_INDEX(`location`, ":", 2), 
  ":", 
  "."
)
```

levimeahan: `SUBSTRING_INDEX()` (nifty function, TIL!) will return the substring up until `count` instances of `delim`, which makes it perfect for returning the first two parts of a location like`20:2:1`

levimeahan (commented):
Looks good! Two potential changes to make but I'll leave those up to you

classes/User.php
AlvinValens: Ah, I left this here. I was going to do a check to verify that the location should be in x:x:x format but the migration should be taking care of this.

db/migrations/20230306004011_travel_overhaul.php
AlvinValens: Bruh, forreals. It's like 3 lines.

PR #275, opened 2023-03-12
Page fixes

HitoriWoB:
Updated dev tools to calculate exp

Added support for HeadAdmins to skip exams

Moved side menu construction to User class
-Menu headers are now generated dynamically
-Added support to limit menu displays correctly
---E.g. travel will no longer show for Rank 1
-Fixed 500 errors on static pages (rank update issue)
-Added support to static pages to show side menu
-Updated support page to use the new sidemenu method
-Removed redundant User Menu headers from layout pages
-Moved logout limit calculation to User class
-Fixed manual to not be broken from class exclusion when not logged in
---Bloodlines have been and remain missing in the manual


HitoriWoB:
Not ready for merging. Need to wait for travel upgrades to be finalized.

Issue #102, opened 2021-10-01
Docker environment for development


tomaszbawor:
Hey, 

Have you considered dockerizing development environment? I would like to contribute (And could take care of setting that up) . Also lot of people are rn using docker on their local machines and that would simplify setting up project?

levimeahan:
Hmm yeah that could be nice - I haven't done any setup work with docker (just used containers), what would be involved?

tomaszbawor:
I have made several things, I want to prepare docker-compose file with setup to start whole application. Basicly after pulling out the project you will be able to run only 'docker-compose up' and it would prepare containers with database (with version as on prod), and container with application exposing ports so no further configuration would be needed.

PR #281, opened 2023-04-04
Add PHP server to the docker-compose setup

drimer:
Addresses https://github.com/elementum-games/shinobi-chronicles/issues/102

I was getting my dev setup, and the most important thing that this addresses for me is better troubleshooting of emails, especially registration for initial testing. I'm sure there are ways of setting that up in the XAMPP installation, but I find it helpful to have it automated through docker-compose.


levimeahan:
Looks good, thanks! I am not very skilled with Docker - One callout I'll make is we're not using Flyway outside of docker for DB migrations, we're using Phinx. Is there an easy way to integrate that with Docker?

drimer:
There should be a way to use Flyway. I'll check it out.

PR #282, opened 2023-04-04
Fix hardcoced react dependencies from unpkg.com

drimer:
Those URLs are returning 404. Bringing them into the node dependencies and importing from there instead.

The unavailability of these resources, which was impacting the game today.

image

I'm not sure about all the changes in the lock file though. Please feel free to edit that if it's introducing any issues.

Worth noting that this may not fix things depending on how you're packaging the dependencies before deploying, but it does fix it locally.


levimeahan:
Hmm, I don't think this is the route we want to go right now. Usually a CDN has a number of advantages for loading these files, being located closer to the user, it has a favorable network path for faster download speed. 

At some point we may bring in a bundler like webpack, but for now I'll move these links to CDNJS which should be more reliable

drimer:
My thinking hehind this change was to remove a 3rd party dependency, even if it'sjust a temporary measure, or a middle step for something else, because these were clearly impacting people when they became unavailable. 

But yes, you make a fair point, a CDN makes way more sense. I'm closing this.

PR #286, opened 2023-04-17
Cast Speed evasion text gender fix.

aumu1031:
BattleManager.parseCombatText now takes into account separate gender values for attacker/target.

Addresses bug where evasion battle text pronoun varies based on which fighter initiates combat.
e.g. [male] attacks [female]: "[female] cast her jutsu before..."
e.g. [female] attacks [male]: "[female] cast his jutsu before..."

### DIFF
--- a/classes/battle/BattleManager.php
+++ b/classes/battle/BattleManager.php
@@ -416,14 +416,18 @@ class BattleManager {
             [
                 '[player]',
                 '[opponent]',
-                '[gender]',
-                '[gender2]'
+                '[attackerGender]',
+                '[attackerGender2]',
+                '[targetGender]',
+                '[targetGender2]'
             ],
             [
                 $attacker->getName(),
                 $target->getName(),
                 $attacker->getSingularPronoun(),
                 $attacker->getPossessivePronoun(),
+                $target->getSingularPronoun(),
+                $target->getPossessivePronoun(),
             ],
             $text
         );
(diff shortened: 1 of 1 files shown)


classes/battle/BattleManager.php
```diff
@@ -416,14 +416,18 @@ private function parseCombatText(string $text, Fighter $attacker, Fighter $targe
             [
                 '[player]',
                 '[opponent]',
-                '[gender]',
-                '[gender2]'
+                '[attackerGender]',
+                '[attackerGender2]',
```
levimeahan, line 420:
```suggestion
                '[gender]',
                '[gender2]',
```
A lot of the jutsu text use `[gender]` for the caster, we'd have to replace all of them, so this is the easy way to ensure backwards compatibility while also allowing for specifying target's gender

classes/battle/BattleManager.php
```diff
@@ -925,7 +929,7 @@ public function jutsuCollision(
                     $collision_text .= "[player] swiftly evaded " . ($damage_reduction * 100) . "% of [opponent]'s damage!";
                 }
                 else {
-                    $collision_text .= "[player] cast [gender2] jutsu before [opponent] cast, negating " .
+                    $collision_text .= "[player] cast [attackerGender2] jutsu before [opponent] cast, negating " .
```
levimeahan, line 932:
```suggestion
                    $collision_text .= "[player] cast [gender2] jutsu before [opponent] cast, negating " .
```
Going along with above

levimeahan (commented):
Thanks for the fix! One change to make this backwards compatible with existing Jutsu text and I think this is good to go

levimeahan (approved):

PR #288, opened 2023-04-19
Apply AK credit from existing premium time when changing seal.

aumu1031:
Converts premium time remaining to AK with following ratios of days->AK rounding down:
Twin Sparrow Seal -> 6:1
Four Dragon Seal -> 2:1

### DIFF
--- a/classes/ForbiddenSeal.php
+++ b/classes/ForbiddenSeal.php
@@ -200,4 +200,18 @@ class ForbiddenSeal {
         $this->journal_image_display = self::$benefits[$this->level]['journal_image_display'];
         $this->avatar_size_display = self::$benefits[$this->level]['avatar_size_display'];
     }
+
+    public function calcRemainingCredit(): int {
+        $akRatio = 1;
+        switch($this->level) {
+            case 1:
+                $akRatio = 6;
+                break;
+            case 2:
+                $akRatio = 2;
+                break;
+            default:
+        }
+        return floor(floor($this->seal_time_remaining / 86400) / $akRatio);
+    }
 }
--- a/classes/ForbiddenSeal.php
+++ b/classes/ForbiddenSeal.php
@@ -202,16 +202,16 @@ class ForbiddenSeal {
     }
 
     public function calcRemainingCredit(): int {
-        $akRatio = 1;
+        $days_per_ak = 1;
         switch($this->level) {
             case 1:
-                $akRatio = 6;
+                $days_per_ak = 6;
                 break;
             case 2:
-                $akRatio = 2;
+                $days_per_ak = 2;
                 break;
             default:
         }
-        return floor(floor($this->seal_time_remaining / 86400) / $akRatio);
+        return floor(floor($this->seal_time_remaining / 86400) / $days_per_ak);
     }
 }
--- a/pages/premium.php
+++ b/pages/premium.php
@@ -581,7 +581,10 @@ function premium() {
                     // Convert remaining premium time to days and calculate AK value
                     $akCredit = $player->forbidden_seal->calcRemainingCredit();
                     // Adjust purchase cost with minimum 0
-                    if (($akCost -= $akCredit) < 0) {$akCost = 0;}
+                    $akCost -= $akCredit;
+                    if ($akCost < 0) {
+                       $akCost = 0;
+                     }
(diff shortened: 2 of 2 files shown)


pages/premium.php
```diff
@@ -578,12 +578,36 @@ function premium() {
                 // Confirm change in seal... time will not be reimbursed
                 if(!isset($_POST['change_forbidden_seal'])) {
                     $confirmation_type = 'change_forbidden_seal';
-                    $confirmation_string = "Are you sure you would like to change from your {$player->forbidden_seal->name}?<br />
-                    You will lose {$system->time_remaining($player->forbidden_seal->seal_time_remaining)} of premium time.<br />
-                    <b>This can not be undone!</b>";
+                    if ($player->forbidden_seal->name == "Twin Sparrow Seal") {
+                        // Convert remaining premium time to days and calculate AK value
+                        $akCredit = floor(floor($player->forbidden_seal->seal_time_remaining / 86400) / 6);
```
HitoriWoB, line 583: Since we have to do this calculation in more than once spot, making it a method would make maintenance easier.
`$akCredit = ForbiddenSeal::calcRemainingCredit();`

pages/premium.php
```diff
@@ -578,12 +578,36 @@ function premium() {
...
+                        // Adjust purchase cost with minimum 0
+                        if (($akCost -= $akCredit) < 0) {$akCost = 0;}
+                        $confirmation_string = "Are you sure you would like to change from your {$player->forbidden_seal->name}?<br />
+                        You will lose {$system->time_remaining($player->forbidden_seal->seal_time_remaining)} of premium time.<br />
+                        Up to {$akCredit} Ancient Kunai will be credited toward your purchase from existing premium time.<br />
+                        <b>This can not be undone!</b>";
+                    }
+                    else if ($player->forbidden_seal->name == "Four Dragon Seal") {
```
HitoriWoB, line 591: Similar to above with Twin seal.

aumu1031: Added second commit simplifying per Hitori's suggestions.

pages/premium.php
```diff
@@ -578,8 +578,13 @@ function premium() {
                 // Confirm change in seal... time will not be reimbursed
                 if(!isset($_POST['change_forbidden_seal'])) {
                     $confirmation_type = 'change_forbidden_seal';
+                    // Convert remaining premium time to days and calculate AK value
+                    $akCredit = $player->forbidden_seal->calcRemainingCredit();
+                    // Adjust purchase cost with minimum 0
+                    if (($akCost -= $akCredit) < 0) {$akCost = 0;}
```
levimeahan, line 584: I know it's valid syntax to do assignments inside if statements, but I'd rather avoid those wherever possible and make the intention very clear
```suggestion
                    $akCost -= $akCredit;
                    if ($akCost < 0) {
                       $akCost = 0;
                     }
```

pages/premium.php
```diff
@@ -595,6 +600,11 @@ function premium() {
                         $message .= " This purchase removed {$system->time_remaining($player->forbidden_seal->seal_time_remaining)}" .
                             " of their {$player->forbidden_seal->name}.";
                     }
+                    // Recalculate adjusted akCost
+                    if ($player->forbidden_seal->level > 0) {
+                        $akCredit = $player->forbidden_seal->calcRemainingCredit();
+                        if (($akCost -= $akCredit) < 0) {$akCost = 0;}
```
levimeahan, line 606: same comment here

classes/ForbiddenSeal.php
```diff
@@ -200,4 +200,18 @@ public function setBenefits() {
         $this->journal_image_display = self::$benefits[$this->level]['journal_image_display'];
         $this->avatar_size_display = self::$benefits[$this->level]['avatar_size_display'];
     }
+
+    public function calcRemainingCredit(): int {
+        $akRatio = 1;
+        switch($this->level) {
+            case 1:
+                $akRatio = 6;
```
levimeahan, line 208: this isn't super intuitive how we ended up at these numbers, but I get this is `<days of seal> / <cost of seal>`. With that said maybe we rename this to `$days_per_ak` to make that more clear?

In a future PR I'd like to then make this based off whatever variable we're basing the cost off of (e.g. move forbidden seal cost definition into this class) so that if we ever change it, this stays up to date. But that's not a blocker for this one 👍

levimeahan (approved):

PR #289, opened 2023-04-25
Fix for inaccurate initial stat transfer cost

aumu1031:
Initializes display text using existing javascript. Also moves inline script after table element.

### DIFF
--- a/templates/premium/premium.php
+++ b/templates/premium/premium.php
@@ -144,40 +144,6 @@
             <th>Stat Transfers</th>
         </tr>
         <tr>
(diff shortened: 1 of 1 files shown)


aumu1031:
Original bug can be replicated with users at Chuunin or higher: observe initial AK cost of stat transfer on page load.

levimeahan (approved):
Hmm I think we also could've just updated the existing initial PHP calculation to use the `$stat_transfer_points_per_ak` variable, but agree this way there's no chance for differences 👍 

LGTM!

PR #291, opened 2023-04-28
Enable cripple effect

aumu1031:
Adds cripple to the `BattleEffectManager.setEffect()` effect list. 
Previously unlisted and defaults to `$apply_effect = false;`

Users report this was working previously, I'm not sure if something else broke at some point but this is the apparent fix.


levimeahan (approved):
Nice catch, thanks for the fix!

PR #293, opened 2023-04-29
Adjust bloodline stat calculations for transparency

aumu1031:
Suggested changes: 
- Simplifies bloodline/offense calculation so that players need only keep bloodline and offense skills equal for optimal stat allocation rather than keeping bloodline stat exactly 110 points ahead. 
- Factors in bloodline speed/cspeed bonus for speed buff/nerf calculations. Bloodline speed bonuses being immutable is interesting but unintuitive. It also creates balance problems, particularly when speed-focused players benefit less from the use of speed boosts.

### DIFF
--- a/classes/battle/BattleEffectsManager.php
+++ b/classes/battle/BattleEffectsManager.php
@@ -211,10 +211,10 @@ class BattleEffectsManager {
             $target->genjutsu_boost += $effect->effect_amount;
         }
         else if($effect->effect == 'cast_speed_boost') {
-            $target->cast_speed_boost += ($target->cast_speed + $target->bloodline_cast_speed_boost) * ($effect->effect_amount / 100);
+            $target->cast_speed_boost += $target->getCastSpeed(true) * ($effect->effect_amount / 100);
         }
         else if($effect->effect == 'speed_boost' or $effect->effect == 'lighten') {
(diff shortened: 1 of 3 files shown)


classes/battle/BattleEffectsManager.php
```diff
@@ -211,10 +211,10 @@ public function applyPassiveEffect(Fighter $target, BattleEffect $effect): bool
             $target->genjutsu_boost += $effect->effect_amount;
         }
         else if($effect->effect == 'cast_speed_boost') {
-            $target->cast_speed_boost += $target->cast_speed * ($effect->effect_amount / 100);
+            $target->cast_speed_boost += ($target->cast_speed + $target->bloodline_cast_speed_boost) * ($effect->effect_amount / 100);
```
levimeahan, line 214: To make this slightly more structured, could we extract this to a `getCastSpeed(bool $include_bloodline = false)` method in the Fighter class, and call it here as `$target->getCastSpeed(true) * ($effect->effect_amount / 100)`?

aumu1031: Agreed, I made the revisions.

levimeahan (approved):
Sweet! Thanks for the fixes

PR #298, opened 2023-05-05
Jutsu page overhaul

aumu1031:
Retains existing equip/learn/forget functionality.
Sample of jutsu view:
image

### DIFF
--- a/pages/jutsu.php
+++ b/pages/jutsu.php
@@ -179,6 +179,8 @@ function jutsu(): void {
 
     $system->printMessage();
     $player->updateInventory();
+    // sort jutsu list by base power
+    usort($player->jutsu, function($a, $b) {return $a->base_power < $b->base_power ? 1 : -1;});
 
     require 'templates/jutsu_page.php';
 }
--- a/pages/jutsu.php
+++ b/pages/jutsu.php
@@ -18,6 +18,8 @@ function jutsu(): void {
 
     $player->getInventory();
 
+    $rank_names = RankManager::fetchNames($system);
+
     $max_equipped_jutsu = User::BASE_JUTSU_SLOTS;
     if($player->rank_num >= 3) {
         $max_equipped_jutsu++;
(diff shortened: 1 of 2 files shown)


pages/jutsu.php
```diff
@@ -164,28 +166,22 @@ function jutsu(): void {
...
-                $jutsu_to_view_child_names[] = $row['name'];
-            }
+            $child_jutsu[$row['parent_jutsu']] = [array($row['name'])];
         }
     }
 
+    $system->printMessage();
+    $player->updateInventory();
+    // sort jutsu list by base power
+    usort($player->jutsu, function($a, $b) {return $a->base_power < $b->base_power ? 1 : -1;});
```
levimeahan, line 183: My only blocking comment on this PR - This will modify the `$player->jutsu` array itself which gets written back to the DB. That won't necessarily cause problems but to avoid any potential unintended consequences I'd rather assign this to another array like `$player_jutsu`, sort that, and use that in the template.

templates/jutsu_page.php
```diff
@@ -1,126 +1,303 @@
...
+        $(".jutsu_select").on('dragleave', function () {
+            this.parentElement.classList.remove('over');
+        });
+        $(".jutsu_select").on('drop', function (e) {
+            e.preventDefault();
+            this.parentElement.classList.remove('over');
+            console.log(e.originalEvent.dataTransfer.getData('text/html'));
+            $(this).val((e.originalEvent.dataTransfer.getData('text/html')));
+        });
+    });
```
levimeahan, line 287: At the point where we have this much display logic ideally would love to see it in a React component or a similar functional pattern where:
- You have a function called `render(currentState)`
- It contains all the display logic to show the whole page based off of whatever `currentState` is
- Then whenever you have an event listener or whatnot you can just update `currentState` and call `render()`

Not a blocker as I prefer to get the UX improvement for users, but directionally would like to move away from large blocks of jQuery manually mutating DOM elements in event handlers as it gets a bit unwieldy to reason around

levimeahan (approved):

PR #299, opened 2023-05-08
Add tooltip to jutsu blocks | Fix display for barrier

aumu1031:


levimeahan (approved):
Sweet! LGTM, thanks for the fix!

PR #300, opened 2023-05-09
Village Reputation addition

HitoriWoB:
Village Reputation introduction (and news post creation link fix).

### DIFF
--- a/classes/Village.php
+++ b/classes/Village.php
@@ -7,6 +7,14 @@ class Village {
     public string $name;
     public string $kage_name;
 
+    const ARENA_REP_GAIN = 1;
+    const PVP_REP_GAIN = 3;
+    const DAILY_REP_GAIN = [
+        DailyTask::DIFFICULTY_EASY => 3,
+        DailyTask::DIFFICULTY_MEDIUM => 6,
+        DailyTask::DIFFICULTY_HARD => 10,
+    ];
+
     public function __construct($system, $village) {
         $this->system = $system;
 
@@ -31,6 +39,9 @@ class Village {
                 return "Kage";
         }
     }
+    public static function getRepName($rep_amount) {
+        return self::repName($rep_amount);
+    }
 
     public function setVillageCoords() {
         return self::getLocation($this->system, $this->name);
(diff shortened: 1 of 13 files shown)


db/migrations/20230509152120_village_reputation.php
```diff
@@ -0,0 +1,25 @@
...
+     * https://book.cakephp.org/phinx/0/en/migrations.html#the-change-method
+     *
+     * Remember to call "create()" or "update()" and NOT "save()" when working
+     * with the Table class.
+     */
+    public function change(): void
+    {
+        $this->execute("ALTER TABLE `users` ADD `village_rep` INT NOT NULL DEFAULT 0;");
+        $this->execute("UPDATE `users` SET `clan_office` = 0");
+        $this->execute("UPDATE `clans` SET `leader`=0, `elder_1`=0, `elder_2`=0");
```
levimeahan, line 23: Let's remove this per not wanting to hard kick current clan officers immediately

classes/Village.php
```diff
@@ -53,4 +64,39 @@ public static function getLocation(System $system, string $village_name): ?Trave
...
+            }
+            elseif($rep_amount >= 200) {
+                return 'Well-known Villager';
+            }
+            elseif($rep_amount >= 100) {
+                return 'Villager';
+            }
+            else {
+                return 'New Villager';
+            }
```
levimeahan, line 95: We reference these tiers elsewhere with hardcoded references to 1000 and 400 reputation - What if we want to change these values? 

Let's make this an array of tiers, and in e.g. the clan officer claim, check the amount and the name for e.g. `reputation_tier[8]` instead of `getRepName(1000)`.

pages/arena.php
```diff
@@ -191,6 +195,44 @@ function processArenaBattleEnd(BattleManager|BattleManagerV2 $battle, User $play
...
+                    $village_rep_chance++;
+                }
+            }
+            elseif($level_difference >= 5) {
+                $village_rep_chance += mt_rand(15, 25);
+            }
+            else {
+                $village_rep_chance += mt_rand(5, 10);
+            }
+        }
```
levimeahan, line 216: Since we're starting reputation from scratch, what if we increased the amount of reputation both granted and required all around so we can give consistent amounts instead of random chance? 

As example for arena this could be like

- 2 or more levels lower = no reputation
- 1 level lower <> 2 levels higher = 1 reputation
- 3+ levels higher = 2 reputation

We'd also need to increase the proportional gains from other sources like PvP and whatnot, as well as the weekly limit and amount required for each reputation tier.

Additionally, we could move this logic to the village class, e.g. a method like`arenaRepGain($player_level, $opponent_level)`

pages/missions.php
```diff
@@ -294,6 +294,43 @@ function runActiveMission(): bool {
                     echo "You have been paid &yen;$mission->money.<br />";
                 }
 
+                // Village reputation
+                // TODO: Consider diminishing returns or daily/weekly caps
+                $reputation_gain = 1;
+                $reputation_chance = 50;
```
levimeahan, line 300: Similar comments here, let's do fixed amounts of reputation instead of random chance

db/migrations/20230509152120_village_reputation.php
HitoriWoB: Will do, hadn't updated that after our discussion. I'll hard code it to be allow ~20 days after launch.

PR #303, opened 2023-05-29
Replace logic for check if sensei, check if sensei on reset

aumu1031:


levimeahan (approved):
Thanks for pushing these through!

PR #305, opened 2023-05-29
Update profile display for student/sensei, display student/sensei in members

aumu1031:


aumu1031:
Added second commit, fixes case where sensei_id is set for sensei prior to resign

levimeahan (approved):

PR #315, opened 2023-06-06
Expand notification functionality, attributes column for data - missions complete

aumu1031:


classes/notification/NotificationAPIManager.php
```diff
@@ -64,6 +64,28 @@ public function getUserNotifications(): array
...
+                case "challenge":
+                    $row['action_url'] = $this->system->router->getUrl("profile");
+                case "team":
+                    $row['action_url'] = $this->system->router->getUrl("profile");
+                case "marriage":
+                    $row['action_url'] = $this->system->router->getUrl("profile");
+                case "student":
+                    $row['action_url'] = $this->system->router->getUrl("profile");
+                case "inbox":
+                    $row['action_url'] = $this->system->router->getUrl("profile");
```
levimeahan, line 88: Should most of these point to the relevant pages? Team, inbox, sensei, etc

classes/notification/NotificationDto.php
```diff
@@ -10,6 +10,7 @@ public function __construct(
         public int $created = 0,
         public int $duration = 0,
         public bool $alert = false,
+        public array $attributes = [],
```
levimeahan, line 13: Instead of an untyped array, could we make this typed? I think everywhere else we have data in an untyped array it becomes more difficult to work with. Some options here include

A) Create subclasses of NotificationDto for each type that needs attributes. For instance `MissionNotificationDto`  that adds on the `mission_id` and `mission_url` fields to the class, internally it gets serialized to/from a JSON string for storage in the database, but anywhere in PHP you work with it, it's typed

B) Create just a subclass for the attributes itself, like `NotificationMissionAttributes`. Make the `attributes` field on this class a union of `null | NotificationMissionAttributes | NotificationSomeOtherAttributes`. So that way when accessing it, we can check `instanceof NotificationMissionAttributes` and then know exactly what we're dealing with and what fields are set

db/migrations/20230605221226_notification_attributes_migration.php
```diff
@@ -0,0 +1,19 @@
...
+{
+    /**
+     * Migrate Up.
+     */
+    public function up(): void
+    {
+        // add new columns
+        $this->execute("
+            ALTER TABLE `notifications`
+                ADD COLUMN `attributes` VARCHAR(200) NOT NULL DEFAULT '[]'
```
levimeahan, line 16:
```suggestion
                ADD COLUMN `attributes` VARCHAR(2000) NOT NULL DEFAULT '[]'
```
Let's make this much larger than we think we need, running into "I stored too much JSON and the DB is truncating it" can be hard to figure out sometimes

levimeahan (approved):
Approved with one blocking comment - Please change the DB migration to have 2000 varchar length instead of 200, since we cannot edit migrations once pushed to the codebase and people start running them.

The other comments can be addressed in a follow up if preferred

classes/notification/NotificationAPIManager.php
aumu1031: Just placeholders, will point to their respective pages.

PR #318, opened 2023-06-07
Add all notification functionality

aumu1031:


classes/notification/types/MissionNotificationDto.php
```diff
@@ -0,0 +1,26 @@
...
+        $notification = new MissionNotificationDto();
+        $notification->type = $row['type'];
+        $notification->message = $row['message'];
+        $notification->notification_id = $row['notification_id'];
+        $notification->user_id = $row['user_id'];
+        $notification->created = $row['created'];
+        $notification->duration = $row['duration'];
+        $notification->alert = $row['alert'];
+        $notification->action_url = $action_url;
+        $notification->mission_rank = $attributes['mission_rank'];
```
levimeahan, line 17: Do we need mission ID too?

index_new.php
```diff
@@ -166,7 +166,9 @@
                 if(strlen($e->getMessage()) > 1) {
                     // Display page title if page is set
                     if($routes[$id] != null) {
-                        echo str_replace("[HEADER_TITLE]", $route->title, $layout->body_start);
+                    require($layout->sidebarModule);
+                    require($layout->topbarModule);
+                    echo str_replace("[HEADER_TITLE]", $route->title, $layout->body_start);
```
levimeahan, line 171: Would be great if we can pull these three lines to a $layout class helper method like `$layout->renderBodyStart(string $page_title)` that can then internally sort out what to require and display

pages/missions.php
```diff
@@ -63,15 +63,30 @@ function missions(): bool {
...
+            else {
+                require_once __DIR__ . '/../classes/notification/NotificationManager.php';
+                $new_notification = new NotificationDto(
+                    type: "mission",
+                    message: $missions[$mission_id]['name'] . " in progress",
+                    user_id: $player->user_id,
+                    created: time(),
+                    attributes: array('mission_rank' => Mission::$rank_names[$missions[$mission_id]['rank']]),
+                    alert: false,
+                );
```
levimeahan, line 87:
```suggestion
                $new_notification = new MissionNotificationDto(
                    type: "mission",
                    message: $missions[$mission_id]['name'] . " in progress",
                    user_id: $player->user_id,
                    created: time(),
                    mission_rank: Mission::$rank_names[$missions[$mission_id]['rank'],
                    alert: false,
                );
```

classes/notification/NotificationDto.php
```diff
@@ -10,7 +12,26 @@ public function __construct(
         public int $created = 0,
         public int $duration = 0,
         public bool $alert = false,
-        public array $attributes = [],
+        protected array $attributes = [],
```
levimeahan, line 15:
```suggestion
```
To go along with previous discussion, I think we should not have this as an actual variable on the NotificationDto class. The `getAttributes()` method below can return a static empty array

classes/notification/types/MissionNotificationDto.php
aumu1031: Just `mission_rank` for the display, other necessary info (mission name, location) passed to message.

index_new.php
aumu1031: Making note for future `index_new` overhaul, most other logic from `index` and `layout` needs replacement.

pages/missions.php
aumu1031: Adjusted latest commit.

levimeahan (approved):
LGTM!

PR #321, opened 2023-06-09
Resist display

Algharna:
Damage resists show in battle, how much is resisted, for normal damage and for residuals. They show up in combat log as well
Centered some pages: view battles and combat logs
image

image
image
image


Algharna:
forgot a display. pulling for now

PR #322, opened 2023-06-09
Fix terms of service new layout

aumu1031:


terms.php
```diff
@@ -1 +1,33 @@
...
+
+if ($layout->key == "new_geisha") {
+    require($layout->headerModule);
+    echo $layout->heading;
+    require($layout->sidebarModule);
+    require($layout->topbarModule);
+    echo "<div id='content'>";
+} else {
+    $layout->renderStaticPageHeader('Terms of Service');
+}
```
levimeahan, line 23: Hmm why not move this logic inside `$layout->renderStaticPageHeader()`?

levimeahan (approved):

terms.php
aumu1031: `$layout->renderStaticPageHeader()` fits here functionally but the the sidebar and topbar aren't conceptually part of a "Static Page Header". For a proper conversion I assume we would want to put those modules into their own functions, but the existing functions for them (e.g. `renderSideMenu()`) are not used in the same order between the old and new layouts.

In the meantime I could modify `$layout->renderStaticPageHeader()` to the following and add necessary System/Player variables as optional parameters:

```
public function renderStaticPageHeader(string $page_title, ?System $system = null, ?User $player = null): void {
        if ($this->key == "new_geisha") {
            require($this->headerModule);
            echo $this->heading;
            require($this->sidebarModule);
            require($this->topbarModule);
            return;
        }
        echo $this->heading;
        echo $this->top_menu;
        echo $this->header;
        echo str_replace("[HEADER_TITLE]", $page_title, $this->body_start);
    }
```

PR #328, opened 2023-06-11
Update training_complete notification (hover), add initial user configuration

aumu1031:


templates/settings.php
```diff
@@ -89,6 +90,24 @@
                     </select>
                     <input type='submit' name='change_layout' value='Change' />
                 </form>
+                <?php if($player->layout == "new_geisha"): ?>
+                <label>Avatar Style</label>
+                <form action='<?=$system->router->getUrl('settings')?>' method='post'>
+                    <select name='avatar_style'>";
+                        <option value='round' <?=($avatar_style == "round" ? "selected='selected'" : "")?>>Round</option>
+                        <option value='four-point' <?=($avatar_style == "four-point" ? "selected='selected'" : "")?>>Four-Point</option>
```
levimeahan, line 98:
```suggestion
                        <option value='four-point' <?=($avatar_style == "four-point" ? "selected='selected'" : "")?>>Square</option>
```
I think people will know this more as "square" although we could also totally include a preview (especially when we add in more)

templates/settings.php
```diff
@@ -89,6 +90,24 @@
...
+                    <input type='submit' name='change_avatar_style' value='Change' />
+                </form>
+                <label>Sidebar Position (Coming Soon)</label>
+                <form action='<?=$system->router->getUrl('settings')?>' method='post'>
+                    <select name='sidebar_position'>";
+                        <option value='left' <?=($sidebar_position == "left" ? "selected='selected'" : "")?>>Left</option>
+                        <option value='right' <?=($sidebar_position == "right" ? "selected='selected'" : "")?>>Right</option>
+                    </select>
+                    <input type='submit' name='change_sidebar_position' value='Change' />
+                </form>
```
levimeahan, line 109: If this is coming soon and not out yet, can we lock behind a `if($system->environment == System::ENVIRONMENT_DEV)` check?

levimeahan (approved):
LGTM 🙌

PR #330, opened 2023-06-12
Add row locking, fix sidebar menu font, inbox

aumu1031:
- Add functions for start/commit/rollback transaction to System
- Add start transaction to all entry after System start
- Add commit to all exit
- Add rollback to all error
- Include System variable to toggle row locking
- Avoid touching specific files for now (register, password, paypal)


api/legacy_special_missions.php
```diff
@@ -6,13 +6,15 @@
...
 $system = new System();
+$system->startTransaction();
 $system->is_api_request = true;
 
 try {
     $player = Auth::getUserFromSession($system);
     $player->loadData();
 } catch(Exception $e) {
     echo json_encode(['logout' => true]);
+    $system->rollbackTransaction();
```
levimeahan, line 17:
```suggestion
    $system->rollbackTransaction();
    error_log($e->getMessage());
```

We should capture these errors in server logs so we can see them

api/notification.php
```diff
@@ -4,14 +4,15 @@
 require_once __DIR__ . "/../classes.php";
 
 $system = new System();
+$system->startTransaction();
```
levimeahan, line 7: Huh, wonder if we should just add this call to the constructor method of System (with a parameter to disable it if desired)

classes/System.php
```diff
@@ -340,6 +343,27 @@ public function db_fetch_all($result = false, ?string $id_column = null): array
...
 
+    public function startTransaction() {
+        if ($this->ENABLE_ROW_LOCKING) {
+            $this->query("START TRANSACTION;");
+            $this->in_db_transaction = true;
+        }
+    }
+
+    public function commitTransaction() {
+        if ($this->ENABLE_ROW_LOCKING) {
```
levimeahan, line 354:
```suggestion
        if ($this->ENABLE_ROW_LOCKING && $this->in_db_transaction) {
```

classes/System.php
```diff
@@ -340,6 +343,27 @@ public function db_fetch_all($result = false, ?string $id_column = null): array
...
+            $this->query("COMMIT;");
+            $this->in_db_transaction = false;
+        }
+    }
+    public function rollbackTransaction()
+    {
+        if ($this->ENABLE_ROW_LOCKING) {
```
levimeahan, line 361:
```suggestion
        if ($this->ENABLE_ROW_LOCKING && $this->in_db_transaction) {
```

manual.php
```diff
@@ -3,6 +3,7 @@
 
 require "classes/System.php";
 $system = new System();
+$system->rollbackTransaction();
```
levimeahan, line 6:
```suggestion
$system->startTransaction();
```

index.php
```diff
@@ -110,6 +111,7 @@
 				throw new Exception("Invalid password! <a href='./password_reset.php'>Forgot password?</a>");
 			}
 		} catch (Exception $e) {
+            $system->rollbackTransaction();
 			$system->message($e->getMessage());
```
levimeahan, line 115:
```suggestion
			$system->message($e->getMessage());
			error_log($e->getMessage());
```

levimeahan: Wonder if we should roll this into one system method, like we have `$system->error` right now, maybe it works or maybe we make a new method that just calls the three lines

```php
$system->rolbackTransaction();
error_log($error_message);
$system->message($error_message);
```

classes/API.php
```diff
@@ -7,7 +7,8 @@
 class API {
 
     #[NoReturn]
-    public static function exitWithError(string $message, array $debug_messages = []): void {
+    public static function exitWithError(string $message, array $debug_messages = [], System $system): void {
+        $system->rollbackTransaction();
```
levimeahan, line 11:
```suggestion
        $system->rollbackTransaction();
        error_log($message);
```

support.php
```diff
@@ -90,6 +91,7 @@
                 }
             }
         }catch (Exception $e) {
+            $system->rollbackTransaction();
```
levimeahan, line 94:
```suggestion
            $system->rollbackTransaction();
            error_log($e->getMessage());
```

support.php
```diff
@@ -157,6 +159,7 @@
                         throw new Exception("Error adding response!");
                     }
                 } catch (Exception $e) {
+                    $system->rollbackTransaction();
```
levimeahan, line 162:
```suggestion
                    $system->rollbackTransaction();
                    error_log($e->getMessage());
```

support.php
```diff
@@ -193,6 +196,7 @@
                         $system->message("Support closed.");
                     }
                 }catch (Exception $e) {
+                    $system->rollbackTransaction();
```
levimeahan, line 199:
```suggestion
                    $system->rollbackTransaction();
                    error_log($e->getMessage());
```

support.php
```diff
@@ -291,6 +295,7 @@
                 $system->message("Error creating support.");
             }
         }catch(Exception $e) {
+            $system->rollbackTransaction();
```
levimeahan, line 298:
```suggestion
            $system->rollbackTransaction();
            error_log($e->getMessage());
```

api/notification.php
aumu1031: We might want to be explicit for visibility since we're being explicit with `$system->commit/rollbackTransaction`?

levimeahan: Yeah, we could leave this for now. But we're pulling the commit into `API::exitWithData`. Maybe the way to go is to make an `API::init` method that returns the $system object with transaction started and `is_api_request` set to true

aumu1031: That approach makes sense. So roll this functionality into the API class:
```
$system = new System();
$system->startTransaction();
$system->is_api_request = true;
return $system;
```
and call this at the start of each API?
`$system = API::init();`

Follow-up PR okay?

levimeahan: Yup, sounds good!

PR #331, opened 2023-06-12
Reputation 2.0

HitoriWoB:


classes/Village.php
```diff
@@ -4,6 +4,78 @@ class Village {
...
+        Mission::RANK_A => 3,
+        Mission::RANK_S => 4
+    ];
+    const DAILY_TASK = [
+        DailyTask::DIFFICULTY_EASY => 1,
+        DailyTask::DIFFICULTY_MEDIUM => 2,
+        DailyTask::DIFFICULTY_HARD => 3,
+    ];
+    const PVP_REP = 3;
+    const DECAY = 0.95;
```
levimeahan, line 77: Clarify here (in comment is fine) what this means - 0.95 rep per day? 5% per day? per week? month?

pages/news.php
```diff
@@ -142,7 +142,7 @@ function news() {
 
 	if(!$page) {
 		if($player->hasAdminPanel()) {
-			echo "<p style='text-align:center;'><a href='$self_link?page=create_post'>New post</a></p>";
+			echo "<p style='text-align:center;'><a href='$self_link&page=create_post'>New post</a></p>";
```
levimeahan, line 145: We finally have a better way to do this, added recently:
```suggestion
			echo "<p style='text-align:center;'><a href='<?= $system->router->getUrl('news', ['page' => 'create_post']) ?>'>New post</a></p>";
```

reputation_cron.php
```diff
@@ -0,0 +1,46 @@
+<?php
```
levimeahan, line 1: We should make a `cron` folder to keep these in one place, since it's a non-obvious part of setting up on a new server that scheduled jobs need to be set up in server config to run scripts like this

levimeahan (approved):
LGTM! Some non-blocking comments if you want to address, otherwise feel free to merge

db/migrations/20230610184101_village_rep.php
```diff
@@ -0,0 +1,26 @@
...
+     *
+     * Write your reversible migrations using this method.
+     *
+     * More information on writing migrations is available here:
+     * https://book.cakephp.org/phinx/0/en/migrations.html#the-change-method
+     *
+     * Remember to call "create()" or "update()" and NOT "save()" when working
+     * with the Table class.
+     */
+    public function change(): void
```
levimeahan, line 19: Can you split this into `up` and `down` methods? The `change` method is for Phinx reversible migration syntax, for manual SQL queries I think we need to define both ourselves.

HitoriWoB:
Migration should be gtg now... assuming I understood the documentation on it :S

PR #332, opened 2023-06-15
Chat 2.0, Map Objectives, API:init(), misc

aumu1031:


classes/Layout.php
aumu1031, line 1: Fixes hotbar position.

classes/chat/ChatManager.php
```diff
@@ -142,6 +150,45 @@ private function fetchPosts(?int $starting_post_id = null, int $max_posts = self
...
+            // Handle Quotes
+            $pattern = "/\[quote:\d+\]/";
+            $has_quote = preg_match_all($pattern, $post->message, $matches);
+            if ($has_quote) {
+                if ($allow_quote) {
+                    foreach ($matches[0] as $match) {
+                        // get chat post contents
+                        $pattern = "/\[quote:(\d+)\]/";
+                        preg_match($pattern, $match, $id);
+                        $quote = $this->fetchPosts($id[1], 1, false);
```
levimeahan, line 162:
```suggestion
                        $quote = $this->fetchPost($id[1]);
```
It gets a little contrived using the batch method to fetch one post, we should add a helper method like this or just inline the query here.

classes/chat/ChatManager.php
```diff
@@ -142,6 +150,45 @@ private function fetchPosts(?int $starting_post_id = null, int $max_posts = self
...
+                        if ($quote) {
+                            // if id match
+                            if ($quote[0]->id == $id[1]) {
+                                // format each entry in $quotes
+                                $formatted_quote = "<div class='quote_container'><a class='chat_user_name " . implode(' ', $quote[0]->user_link_class_names) . "' href='" . $this->system->router->getURL("members", ["user" => $quote[0]->user_name]) . "'>" . $quote[0]->user_name . "</a><div class='quote_message'>" . $quote[0]->message . "</div></div>";
+                                // run replace on $test using $matches as search and $quotes as replace
```
levimeahan, line 168: This is okay for now but I think we should move more in the direction of embedding display logic in the front end, detecting in the UI component when there's a `quote` tag and replacing it with our desired markup. 

We should also probably drop the current BBcode system and migrate to a completely front-end system using something like Markdown https://github.com/remarkjs/react-markdown, so we can take care of it then.

classes/chat/ChatManager.php
```diff
@@ -142,6 +150,45 @@ private function fetchPosts(?int $starting_post_id = null, int $max_posts = self
...
+                        }
+                    }
+                }
+                else {
+                    $post->message = str_replace($matches[0], "(...)", $post->message);
+                }
+            }
+
+            // Handle Mention
+            $pattern = "/@([^ \n\s]+)(?=[^A-Za-z0-9_]|$)/";
```
levimeahan, line 183: Have you tested which cases stop the username? I'm not sure if it's a regex101 problem but it suggests that non-word characters in the middle of the string would be included
image
image

classes/notification/NotificationAPIManager.php
```diff
@@ -163,6 +163,13 @@ public function getUserNotifications(): array
                         $notifications[] = NotificationDto::fromDb($row, $this->system->router->getUrl("inbox"));
                     }
                     break;
+                case "chat":
+                    if (false) {
```
levimeahan, line 167: Is this supposed to be some other condition, or are we saving for later?

classes/travel/ObjectiveLocation.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+class ObjectiveLocation {
```
levimeahan, line 3: What do you think about naming this `MapObjectiveLocation` just for slightly more clarity?

ui_components/src/chat/Chat_new.js
```diff
@@ -0,0 +1,385 @@
...
+    const [showMemeSelect, setShowMemeSelect] = React.useState(false);
+
+    /*$(document).on("click", ".meme_select", function () {
+        // Chat.val(Chat.val() + $(this).attr("data-code"));
+        $("#meme_modal").addClass("hidden");
+        $("#meme_toggle").text("+ Meme");
+    });
+    $(document).on("click", ".meme_toggle", function (e) {
+        $("#meme_modal").toggleClass("hidden");
+    });*/
```
levimeahan, line 203:
```suggestion
```

ui_components/src/chat/Chat_new.js
```diff
@@ -0,0 +1,385 @@
...
+    isModerator,
+    deletePost,
+    quotePost,
+    goToPreviousPage,
+    goToNextPage
+}) {
+    return (
+        <>
+            <div id="chat_navigation">
+                <div className="chat_navigation_divider_left"><svg width="100%" height="2"><line x1="0%" y1="1" x2="100%" y2="1" stroke="#4e4535" strokeWidth="1"></line></svg></div>
```
levimeahan, line 305:
```suggestion
                <div className="chat_navigation_divider_left">
                  <svg width="100%" height="2"><line x1="0%" y1="1" x2="100%" y2="1" stroke="#4e4535" strokeWidth="1"></line></svg>
                </div>
```

levimeahan: Style-wise, would be great to break these lines up a little bit between the container and contents

ui_components/src/chat/Chat_new.js
```diff
@@ -0,0 +1,385 @@
...
+                                    <a className='imageLink' href={post.reportLink}>
+                                        <img className='post_report' src='../images/v2/icons/report_hover.png' />
+                                    </a>
+                                </div>
+                            </div>
+                            <div className="post_timestamp">
+                                <div style={{ marginBottom: "2px" }}>{post.timeString}</div>
+                            </div>
+                        </div>
+                    </div>
```
levimeahan, line 378: This should all get split out into a `ChatPost` component

levimeahan (approved):

classes/notification/NotificationAPIManager.php
aumu1031: Yeah, there's no default case for when a chat message should be deleted so keeping this a placeholder.

In the future we may decide to use the time created or set duration and delete if too old?

classes/chat/ChatManager.php
aumu1031: Fixed in commit.

classes/travel/ObjectiveLocation.php
aumu1031: Changed in commit.

ui_components/src/chat/Chat_new.js
aumu1031: Marking as to-do.

classes/chat/ChatManager.php
aumu1031: Marking as to-do.

classes/chat/ChatManager.php
aumu1031: Marking as to-do.

aumu1031:
To-do:
`fetchPosts` -> `fetchPost`
Shift quote logic from server to front-end (react-markdown potentially)
Move chat post js into `ChatPost` component

PR #333, opened 2023-06-15
Bloodline resist display

Algharna:
Damage resists from bloodline now shows in battle, how much is resisted, for normal damage and for residuals, this displays in combat log as well.
Center aligned some pages: view battles, combat logs, gear
In battle:
image
Battle log
image


levimeahan (approved):
LGTM! Thanks for doing this

PR #335, opened 2023-06-16
Add new avatar styles, fix chat name/border styles

aumu1031:


ui_components/src/chat/Chat_new.js
```diff
@@ -27,20 +28,23 @@ type Props = {|
...
 }: Props) {
     const [banInfo, setBanInfo] = React.useState(initialBanInfo);
     const [posts, setPosts] = React.useState(initialPosts);
 
     const [nextPagePostId, setNextPagePostId] = React.useState(initialNextPagePostId);
     const [latestPostId, setLatestPostId] = React.useState(initialLatestPostId);
     // Only set if we're paginating
-    const [previousPagePostId, setPreviousPagePostId] = React.useState(null);
+    const [previousPagePostId, setPreviousPagePostId] = React.useState(initialPreviousPagePostId);
+    const initialPostIdRef = React.useRef(initialPostId);
```
levimeahan, line 47: Hmm is this doing something functionally different than `currentPagePostIdRef`, or could we initialize `currentPagePostIdRef` to that value and drop this var completely?

aumu1031: `initialPostIdRef` gives us the option to target the requested post for things like highlighting.

Also if we initialize `currentPagePostIdRef` with the current logic then `refreshChat` won't run when you view latest results.

```
const refreshChat = function() {
        if(currentPagePostIdRef.current != null) {
            return;
        }```

Comment (levimeahan) [reply ui_components/src/chat/Chat_new.js:47]: > initialPostIdRef gives us the option to target the requested post for things like highlighting.
Makes sense. I think what we could do to make this clearer is create a state variable called `highlightPostId` and initialize it to the value of `initialPostId`. 

> Also if we initialize currentPagePostIdRef with the current logic then refreshChat won't run when you view latest results.
Calling `changePage(latestPostId)` will actually clear the `currentPagePostIdRef`, same way as continually clicking "newer" until you hit the latest post clears it

File: templates/chat.php [hunk 2]
```diff
@@ -28,7 +29,11 @@
             React.createElement(Chat, {
                 chatApiLink: "<?= $system->router->api_links['chat'] ?>",
                 initialPosts: initialPostsResponse.posts,
+                <?php if (isset($initialChatPostId)): ?>
+                initialPostId: <?= $initialChatPostId ?>,
+                <?php endif; ?>
```
levimeahan, line 34:
```suggestion
                initialPostId: <?= $initialChatPostId ?? null ?>,
```

ui_components/src/chat/Chat_new.js
```diff
@@ -56,8 +60,11 @@ function Chat({
     }
 
     const refreshChat = function() {
-        if(currentPagePostIdRef.current != null) {
-            return;
+        if (currentPagePostIdRef.current != null) {
+            // always refresh when initialized to latest
+            if (currentPagePostIdRef.current != latestPostId) {
+                return;
+            }
```
levimeahan, line 67:
```suggestion
        return;
```

ui_components/src/chat/Chat_new.js
```diff
@@ -162,6 +169,8 @@ function Chat({
                 quotePost={quotePost}
                 goToNextPage={() => changePage(nextPagePostId)}
                 goToPreviousPage={() => changePage(previousPagePostId)}
+                goToLatestPage={() => changePage(latestPostId)}
+                postsBehind={currentPagePostIdRef.current > 0 ? latestPostId - currentPagePostIdRef.current : 0}
```
levimeahan, line 173:
```suggestion
                postsBehind={currentPagePostIdRef.current != null ? latestPostId - currentPagePostIdRef.current : 0}
```

levimeahan (approved):

PR #337, opened 2023-06-17
Fix Chuunin Exam, misc fixes

aumu1031:
- Chuunin exam (remove deleted reference)
- Add training notification to ancient market
- Fix case on link to latest post
- Fix training display if timer goes negative
- Remove map location for "search" missions
- Fix mission notification display
- Prevent joining team mission if already on mission

### DIFF
--- a/pages/team.php
+++ b/pages/team.php
@@ -176,13 +176,17 @@ function team() {
     }
     else {
         if(isset($_GET['join_mission']) && $player->team->mission_id) {
-            $mission_id = $player->team->mission_id;
-            $mission = new Mission($mission_id, $player, $player->team);
+            if ($player->mission_id) {
+                $system->message("You are already on a mission!");
+            } else {
+                $mission_id = $player->team->mission_id;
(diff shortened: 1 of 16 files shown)


pages/team.php
```diff
@@ -363,6 +397,33 @@ function team() {
                     $system->query("UPDATE `teams` SET `mission_id`='{$mission_id}' WHERE `team_id`='{$player->team->id}' LIMIT 1");
                     if($system->db_last_affected_rows) {
                         $system->message("Mission started!");
+                        // Create notification
+                        $result = $system->query("SELECT `name` FROM `missions` WHERE `mission_id` = {$mission_id}");
+                        $mission_name = $system->db_fetch($result)['name'];
+                        require_once __DIR__ . '/../classes/notification/NotificationManager.php';
```
levimeahan, line 403: Per convention, let's move this require up to the top of the page (difference with this vs layout/templates is that requiring the NotificationManager does not trigger any effects, just makes the class available)

levimeahan (approved):

PR #339, opened 2023-06-19
Sensei tweaks, Training Feature

aumu1031:
- Sensei are considered specialized in their bloodline
- Fix cast speed specialization display
- Sensei set to inactive instead of deleted to preserve data

### DIFF
--- a/classes/SenseiManager.php
+++ b/classes/SenseiManager.php
@@ -10,6 +10,8 @@ class SenseiManager {
 
     public static array $exam_answers = ['1a', '2c', '3c', '4c', '5b', '6b'];
 
+    const LESSON_COST_PER_MINUTE = 60;
+
     public static function scoreExam(array $answers, System $system): bool {
         $passed = true;
         foreach ($answers as $answer) {
@@ -47,26 +49,35 @@ class SenseiManager {
     public static function getSenseiByVillage(string $village_name, System $system): array {
         $sensei_list = [];
         $sensei_table_result = $system->db->query(
-            "SELECT `sensei`.`sensei_id`, `students`, `recruitment_message`, `graduated_count`, `specialization`, `avatar_link`, `user_name`, `user_bloodlines`.`name` FROM `sensei`
+            "SELECT `sensei`.`sensei_id`, `students`, `temp_students`, `recruitment_message`, `graduated_count`, `specialization`, `enable_lessons`, `avatar_link`, `user_name`, `ninjutsu_skill`, `taijutsu_skill`, `genjutsu_skill`, `bloodline_skill`, `speed`, `cast_speed`, `exp` FROM `sensei`
             INNER JOIN `users` ON `sensei`.`sensei_id` = `users`.`user_id`
-            LEFT JOIN `user_bloodlines` ON `sensei`.`sensei_id` = `user_bloodlines`.`user_id`
             WHERE `village` = '{$village_name}' AND `accept_students` = '1' AND `is_active` = 1 ORDER BY `graduated_count` DESC"
         );
         while ($result = $system->db->fetch($sensei_table_result)) {
             $sensei_list[] = [
                 'sensei_id' => $result['sensei_id'],
                 'students' => json_decode($result['students']),
+                'temp_students' => json_decode($result['temp_students'], true),
                 'recruitment_message' => $result['recruitment_message'],
                 'graduated' => $result['graduated_count'],
                 'specialization' => $result['specialization'],
+                'enable_lessons' => $result['enable_lessons'],
 			    'avatar_link' => $result['avatar_link'],
 			    'user_name' => $result['user_name'],
-			    'bloodline_name' => $result['name']
+                'ninjutsu_skill' => $result['ninjutsu_skill'],
+                'taijutsu_skill' => $result['taijutsu_skill'],
+                'genjutsu_skill' => $result['genjutsu_skill'],
+                'bloodline_skill' => $result['bloodline_skill'],
+                'speed' => $result['speed'],
+                'cast_speed' => $result['cast_speed'],
+                'exp' => $result['exp'],
             ];
         }
 		foreach ($sensei_list as &$sensei) {
-            $sensei += SenseiManager::getStudentBoost($sensei['graduated'], $system);
+            $sensei += SenseiManager::getStudentBoostBySensei($sensei['sensei_id'], $system);
 			$sensei['students'] = SenseiManager::getStudentData($sensei['students'], $system);
+            $sensei['temp_students'] = SenseiManager::checkTempStudents($sensei['sensei_id'], $system);
+            $sensei['temp_students'] = SenseiManager::getTempStudentData($sensei['temp_students'], $system);
         }
         return $sensei_list;
     }
@@ -74,16 +85,22 @@ class SenseiManager {
     public static function getSenseiByID(int $sensei_id, System $system): array {
         $sensei = [];
         $sensei_table_result = $system->db->query(
-            "SELECT `students`, `recruitment_message`, `student_message`, `graduated_count`, `specialization` FROM `sensei` WHERE `sensei_id` = '{$sensei_id}'"
+            "SELECT * FROM `sensei` WHERE `sensei_id` = '{$sensei_id}'"
         );
         $result = $system->db->fetch($sensei_table_result);
         $sensei = array(
             'sensei_id' => $sensei_id,
             'students' => json_decode($result['students']),
             'recruitment_message' => $result['recruitment_message'],
-            'student_message' => $result['student_message'],
             'graduated' => $result['graduated_count'],
-            'specialization' => $result['specialization']
+            'student_message' => $result['student_message'],
+            'specialization' => $result['specialization'],
+            'graduated_students' => json_decode($result['graduated_students']),
+            'temp_students' => json_decode($result['temp_students'], true),
+            'time_trained' => $result['time_trained'],
+            'yen_gained' => $result['yen_gained'],
+            'is_active' => $result['is_active'],
+            'enable_lessons' => $result['enable_lessons'],
         );
         return $sensei;
     }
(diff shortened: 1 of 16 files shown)


classes/SenseiManager.php
```diff
@@ -32,67 +34,99 @@ public static function getStudentBoost(int $graduated_count): array {
...
             'graduated' => $result['graduated_count'],
-            'specialization' => $result['specialization']
+            'student_message' => $result['student_message'],
+            'specialization' => $result['specialization'],
+            'graduated_students' => json_decode($result['graduated_students']),
+            'temp_students' => json_decode($result['temp_students'], true),
+            'time_trained' => $result['time_trained'],
+            'yen_gained' => $result['yen_gained'],
+            'is_active' => $result['is_active'],
+            'enable_lessons' => $result['enable_lessons'],
```
levimeahan, line 103: We should make a `SenseiInfoDto` for this, much easier to work with outside this function

classes/SenseiManager.php
```diff
@@ -158,6 +192,17 @@ public static function updateStudentRecruitment(int $sensei_id, $recruitment_mes
...
+    public static function updateStudentLessons(int $sensei_id, bool $enable_lessons, System $system): bool
+    {
+        $db_modified = false;
+        $system->db->query(
+            "UPDATE `sensei` SET `enable_lessons` = '{$enable_lessons}' WHERE `sensei_id` = '{$sensei_id}'"
+        );
+        if ($system->db->last_affected_rows > 0) {
+            $db_modified = true;
+        }
+        return $db_modified;
```
levimeahan, line 204:
```suggestion
        $system->db->query(
            "UPDATE `sensei` SET `enable_lessons` = '{$enable_lessons}' WHERE `sensei_id` = '{$sensei_id}'"
        );
        
        return $system->db->last_affected_rows > 0;
```

classes/SenseiManager.php
```diff
@@ -333,4 +380,120 @@ public static function isSensei(int $user_id, System $system): bool {
...
+    public static function getLessonCostForPlayer(User $player, System $system): array {
+        $cost = [];
+        $stat_train_length = 600;
+	    // 56.25% of standard
+	    $stat_long_train_length = $stat_train_length * 4;
+        // 30x length (5 hrs), 12x gains: 40% of standard
+        $stat_extended_train_length = $stat_train_length * 30;
+	    // Forbidden seal trainings boost
+        $stat_long_train_length *= $player->forbidden_seal->long_training_time;
+        $stat_extended_train_length = round($stat_extended_train_length * $player->forbidden_seal->extended_training_time);
```
levimeahan, line 420: We should not redeclare these values here or it will get out of date if we tweak training page, can we pull this to a method on the `User` class? Or create a shell of a `TrainingManager` class (in future to be used via API, but for now just to hold this "how are trainings setup" logic)

classes/SenseiManager.php
```diff
@@ -333,4 +380,120 @@ public static function isSensei(int $user_id, System $system): bool {
...
+    public static function getLessonDurationForPlayer(User $player, System $system): array {
+        $duration = [];
+        $stat_train_length = 600;
+        // 56.25% of standard
+        $stat_long_train_length = $stat_train_length * 4;
+        // 30x length (5 hrs), 12x gains: 40% of standard
+        $stat_extended_train_length = $stat_train_length * 30;
+        // Forbidden seal trainings boost
+        $stat_long_train_length *= $player->forbidden_seal->long_training_time;
+        $stat_extended_train_length = round($stat_extended_train_length * $player->forbidden_seal->extended_training_time);
```
levimeahan, line 440: same comment, reference from a single source of truth instead of redeclaring here

classes/SenseiManager.php
```diff
@@ -333,4 +380,120 @@ public static function isSensei(int $user_id, System $system): bool {
...
+
+    public static function logLesson(int $sensei_id, int $temp_student_id, int $lesson_duration, array $temp_students, int $yen_gained, int $time_trained, System $system): bool {
+        $db_modified = false;
+        $temp_students[$temp_student_id] = time() + $lesson_duration;
+        $temp_students = json_encode($temp_students, JSON_FORCE_OBJECT);
+        $system->db->query("UPDATE `sensei` SET `temp_students` = '{$temp_students}', `yen_gained` = {$yen_gained}, `time_trained` = {$time_trained} WHERE `sensei_id` = '{$sensei_id}'");
+        if ($system->db->last_affected_rows > 0) {
+            $db_modified = true;
+        }
+        return $db_modified;
```
levimeahan, line 465: I generally lean towards being verbose and explicit but I wonder if this provides that much benefit. If anything maybe we add a comment here, but IMO the logic is straightforward enough to read
```suggestion
        return $system->db->last_affected_rows > 0;
```

pages/premium.php
```diff
@@ -108,13 +110,13 @@ function premium(): bool {
 	if(isset($_POST['user_reset'])) {
 		try {
             if($player->team) {
-                throw new RuntimeException("You must leave your team before resetting!");
+                throw new Exception("You must leave your team before resetting!");
```
levimeahan, line 113:
```suggestion
                throw new RuntimeException("You must leave your team before resetting!");
```
This looks like bad merge conflict, these should all be RuntimeException now

pages/training.php
```diff
@@ -142,9 +144,13 @@ function training() {
 			// Check for sensei training boost
 			if($player->sensei_id != 0) {
 				$sensei_boost = SenseiManager::getStudentBoostBySensei($player->sensei_id, $system);
-				if ($train_type == $sensei_boost['specialization'] . "_skill" || $train_type == $sensei_boost['specialization']) {
+                if ($train_type == $sensei_boost['specialization'] . "_skill" || $train_type == $sensei_boost['specialization']) {
```
levimeahan, line 147: Maybe we could simplify this to a `SenseiManager::getTrainingBoostForTrainType($player->sensei_id, $train_type, $system)`

which in most cases returns 0 or null, but in these two cases returns `$sensei_boost['boost_primary']`

pages/villageHQ.php
```diff
@@ -275,16 +277,16 @@ function villageHQ() {
 		if (isset($_POST['submit_exam'])) {
 			try {
 				// check if already sensei
-				if (SenseiManager::isSensei($player->sensei_id, $system)) {
-                    throw new RuntimeException('You do not meet the requirements!');
+				if (SenseiManager::isActiveSensei($player->sensei_id, $system)) {
+                    throw new Exception('You do not meet the requirements!');
```
levimeahan, line 281: More merge conflicts here
```suggestion
                    throw new RuntimeException('You do not meet the requirements!');
```

pages/training.php
aumu1031: Modified in commit.

db/migrations/20230620062917_sensei_lesson_migration.php
```diff
@@ -0,0 +1,23 @@
...
+        // Modify sensei table
+        $this->execute("
+            ALTER TABLE `sensei`
+                ADD `temp_students` VARCHAR(50) NOT NULL DEFAULT '[]' ,
+                ADD `time_trained` INT(13) NOT NULL DEFAULT '0',
+                ADD `yen_gained` INT(11) NOT NULL DEFAULT '0'`
+                ADD `is_active` TINYINT(1) NOT NULL DEFAULT '1'
+                ADD `enable_lessons` TINYINT(1) NOT NULL DEFAULT '1';
+        ");
+    }
```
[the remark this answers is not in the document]
aumu1031: Fixed migration in commit, had other issues.

classes/SenseiManager.php
aumu1031: Agreed. I'm thinking this is a to-do as part of cleaning up the sensei functionality and converting to API.

pages/training.php
aumu1031: Changed in latest commit.

classes/SenseiManager.php
aumu1031: As part of the hotbar implementation I assume we're moving training, ramen and special missions into APIs? 
So as part of that the training logic on the Training/Sensei/Premium pages will be unified.

pages/villageHQ.php
```diff
@@ -442,32 +448,204 @@ function villageHQ() {
...
+
+				// Set training for player
+				$stat_train_gain = 4 + ($player->rank_num * 4);
+				$stat_long_train_gain = $stat_train_gain * 2.25;
+                $stat_extended_train_gain = $stat_train_gain * 12;
+				$stat_long_train_gain *= $player->forbidden_seal->long_training_gains;
+                $stat_extended_train_gain = round($stat_extended_train_gain * $player->forbidden_seal->extended_training_gains);
+                $stat_train_gain += $system->TRAIN_BOOST;
+                $stat_long_train_gain += $system->LONG_TRAIN_BOOST;
+                $stat_extended_train_gain += ($system->LONG_TRAIN_BOOST * 5);
```
levimeahan, line 589: Tbh I would really prefer we centralize these in the PR that they are introduced (currently, nowhere outside of `training.php` attempts to define these values) rather than introducing them and then cleaning up later. Would be okay if it's in total shell of a `TrainingManager` class which is only used (for now) for exposing these values to wherever needs them.

But I am okay with that being an immediate fast follow

levimeahan (approved):
Approving to unblock, but two pressing comments need to be addressed ASAP (fixing exceptions from the merge conflict is most important)

classes/SenseiManager.php
aumu1031: Moved to TrainingManager

classes/SenseiManager.php
aumu1031: Moved to TrainingManager

pages/villageHQ.php
```diff
@@ -275,16 +277,16 @@ function villageHQ() {
...
 			try {
 				// check if already sensei
-				if (SenseiManager::isSensei($player->sensei_id, $system)) {
+				if (SenseiManager::isActiveSensei($player->sensei_id, $system)) {
                     throw new RuntimeException('You do not meet the requirements!');
                 }
 				// check rank
                 if ($player->rank_num < 4) {
-                    throw new RuntimeException('You do not meet the requirements!');
+                    throw new Exception('You do not meet the requirements!');
```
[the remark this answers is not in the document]
aumu1031: Updated Exceptions -> RuntimeException

pages/villageHQ.php
aumu1031: Moved to TrainingManager

PR #340, opened 2023-06-23
Home page initial commit (dev-only)

aumu1031:
- Add new Home page (login complete)
- Combine Login/Register/Reset to Index
- News API


classes/news/NewsAPIManager.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+require __DIR__ . '/NewsPostDto.php';
+
+class NewsAPIManager {
```
levimeahan, line 5:
```suggestion
class NewsManager {
```
I realize I didn't present this consistently, but I don't think we should include `API` in these manager classes, it's overly specific. We want to use them for anything, not just APIs (as opposed to the presenter and response which specifically are only useful for constructing a response)

index.php
```diff
@@ -47,80 +58,344 @@
...
+            $system->message($e->getMessage());
+            error_log($e->getMessage());
+            $register_error_text = $e->getMessage();
+
+            $register_pre_fill['user_name'] = isset($_POST['user_name']) ? $_POST['user_name'] : "";
+            $register_pre_fill['email'] = isset($_POST['email']) ? $_POST['email'] : "";
+            $register_pre_fill['gender'] = isset($_POST['gender']) ? $_POST['gender'] : "";
+            $register_pre_fill['village'] = isset($_POST['village']) ? $_POST['village'] : "";
+        }
+    }
```
levimeahan, line 398: I want to extract these huge blocks of logic out so `index` can be lighter-weight, but I'll take that as a follow-up for me to do

ui_components/src/home/Home.js
```diff
@@ -0,0 +1,564 @@
+import { apiFetch } from "../utils/network.js";
```
levimeahan, line 1:
```suggestion
// @flow
import { apiFetch } from "../utils/network.js";
```

ui_components/src/home/Home.js
```diff
@@ -0,0 +1,564 @@
...
+        setDisplayRegister(false);
+        setDisplayReset(false);
+        setDisplayRules(false);
+        setDisplayTerms(false);
+        if (!displayLogin) {
+            setDisplayLogin(true);
+        }
+        else {
+            document.getElementById('login_form').submit();
+        }
```
levimeahan, line 96: Looking through these, it doesn't look like there's ever a case we want to have multiple of these be true. What if we moved this to a single `pageToDisplay` state variable, that is a union of `"register" | "reset" | "rules" | "terms" | "login"`

ui_components/src/home/Home.js
```diff
@@ -0,0 +1,564 @@
...
+    loginMessageText,
+    registerPreFill,
+    initialNewsPosts,
+}: Props) {
+    const [displayLogin, setDisplayLogin] = React.useState((loginErrorText != "" || loginMessageText != "") ? true : false);
+    const [displayRegister, setDisplayRegister] = React.useState(registerErrorText == "" ? false : true);
+    const [displayReset, setDisplayReset] = React.useState(resetErrorText == "" ? false : true);
+    const [displayRules, setDisplayRules] = React.useState(false);
+    const [displayTerms, setDisplayTerms] = React.useState(false);
+    const [newsPosts, setNewsPosts] = React.useState(initialNewsPosts);
```
levimeahan, line 28: None of these are used in this component itself - Could we move them down to the `LoginSection` component and avoid a bunch of extra props?

ui_components/src/home/Home.js
```diff
@@ -0,0 +1,564 @@
...
+                                </div>
+                            }
+                            {loginErrorText != "" &&
+                                <div className="login_input_bottom">
+                                    <div className="login_error_label">{loginErrorText}</div>
+                                    <div className="reset_link" onClick={() => handleDisplayReset()}>reset password</div>
+                                </div>
+                            }
+                            </form>
+                        }
```
levimeahan, line 169: If we go the route of having a `pageToDisplay` var, we could extract each of these to its own component and refactor this to be a switch statement something like

```js
switch(pageToDisplay) {
  case "register":
    return <RegisterForm /* props here */ />;
  case "login":
    return <LoginForm /* props here */ />;
  // ...
}
```

levimeahan (approved):

classes/news/NewsAPIManager.php
```diff
@@ -0,0 +1,28 @@
...
+    public function getLatestPosts(int $max_posts = 8) : array {
+        $return_arr = [];
+
+        $result = $this->system->db->query("SELECT * FROM `news_posts` ORDER BY `post_id` DESC LIMIT $max_posts");
+
+        while ($post = $this->system->db->fetch($result)) {
+            $return_arr[] = NewsPostDto::fromDb($post);
+        }
+
+        return $return_arr;
```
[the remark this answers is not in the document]
aumu1031: I'm learning!

api/news.php
```diff
@@ -0,0 +1,45 @@
...
+
+try {
+    // api requires a request
+    if (isset($_POST['request'])) {
+        $request = filter_input(INPUT_POST, 'request', FILTER_SANITIZE_FULL_SPECIAL_CHARS);
+    } else {
+        throw new RuntimeException('No request was made!');
+    }
+
+    $NewsAPIResponse = new NewsAPIResponse();
```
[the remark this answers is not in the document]
aumu1031: I agree that there's no functional value right now - but my thinking is that adding a custom response for each is trivial to do when setting up the API and keeps things standardized between them.

classes/Layout.php
```diff
@@ -20,8 +20,12 @@ public function __construct(
         public string $footer,
     ) {}
 
-    public function renderBeforeContentHTML(System $system, ?User $player, string $page_title): void {
+    public function renderBeforeContentHTML(System $system, ?User $player, string $page_title, bool $custom_page = false): void {
```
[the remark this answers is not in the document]
aumu1031: This is a flag to avoid rendering the layout modules, so header/sidebar/footer, and let the page create those.

classes/System.php
```diff
@@ -97,7 +97,7 @@ class System {
     );
 
     // Default layout
-    const DEFAULT_LAYOUT = 'shadow_ribbon';
+    const DEFAULT_LAYOUT = 'new_geisha';
```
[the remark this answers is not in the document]
aumu1031: Oops. Fixed in commit. 
This is a leftover before an override was added for default layout.

classes/news/NewsAPIManager.php
aumu1031: Fixed in commit.

ui_components/src/home/Home.js
aumu1031: Marking as to-do.

ui_components/src/home/Home.js
aumu1031: Marking as to-do.

ui_components/src/home/Home.js
aumu1031: Marking as to-do.

classes/Layout.php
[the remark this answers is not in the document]
levimeahan: Gotcha - What if we named it something more descriptive of that, like `$content_only` or something?

api/news.php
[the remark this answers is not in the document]
levimeahan: Good point. Let's keep it!

PR #344, opened 2023-06-26
Fix merge, Battle History to ViewBattles, Academy Page, Layout Config

aumu1031:


classes/Layout.php
```diff
@@ -20,32 +20,45 @@ public function __construct(
         public string $footer,
     ) {}
 
-    public function renderBeforeContentHTML(System $system, ?User $player, string $page_title, bool $custom_page = false): void {
+    public function renderBeforeContentHTML(System $system, ?User $player, string $page_title, bool $render_header = true, bool $render_sidebar = true, bool $render_topbar = true, bool $render_content = true): void {
```
levimeahan, line 23: There's enough params we could move to each on its own line for readability
```suggestion
    public function renderBeforeContentHTML(
      System $system, 
      ?User $player, 
      string $page_title, 
      bool $render_header = true, 
      bool $render_sidebar = true, 
      bool $render_topbar = true, 
      bool $render_content = true
    ): void {
```

classes/Layout.php
```diff
@@ -71,9 +84,9 @@ public function renderBeforeContentHTML(System $system, ?User $player, string $p
         }
     }
 
-    public function renderAfterContentHTML(System $system, ?User $player, ?float $page_load_time = null, bool $custom_page = false): void {
+    public function renderAfterContentHTML(System $system, ?User $player, ?float $page_load_time = null, bool $render_content = true, bool $render_footer = true, bool $render_hotbar = true): void {
         if($this->key == 'new_geisha') {
-            if ($custom_page) {
+            if (!$render_content) {
```
levimeahan, line 89: this is way easier to understand at first glance, thank you!

pages/academy.php
```diff
@@ -0,0 +1,499 @@
...
+                    break;
+                case 'bloodline_skill':
+					if ($player->bloodline_id == $lesson_sensei['bloodline_id']) {
+						$lesson_modifier = round(SenseiManager::getLessonModifier($player->bloodline_skill, $player->exp, $lesson_sensei['bloodline_skill'], $lesson_sensei['exp'], true) * 100 - 100, 2);
+					} else {
+						throw new RuntimeException('Invalid stat!');
+                    }
+                    break;
+                case 'speed':
+                    $lesson_modifier = SenseiManager::getLessonModifier($player->speed, $player->exp, $lesson_sensei['speed'], $lesson_sensei['exp'], $lesson_sensei['specialization'] == 'speed' ? true : false);
```
levimeahan, line 294: Totally optional, but we could be slightly more specific by calling this `$lesson_gain_multiplier`  or similar

pages/viewBattles.php
```diff
@@ -106,5 +106,115 @@ function viewBattles() {
     }
 
     require 'templates/viewBattles.php';
+
+    /* Begin User Battle History */
+    if ($player->rank_num > 1 && $player->forbidden_seal->max_battle_history_view > 0) {
```
levimeahan, line 111: Should we make this like, a tab players can switch between? Like

"Live Battles" | "Your Battle History"

levimeahan (approved):
LGTM 👍

pages/viewBattles.php
aumu1031: I think that's a great improvement, will add to the list.

pages/academy.php
aumu1031: Sounds good to me.

PR #346, opened 2023-06-28
New profile skeleton

levimeahan:
Reusable character avatar component is here, also finally got around to making a `$system->isDevEnvironment()` helper method

Progress so far:
image

### DIFF
--- a/ui_components/src/profile/Profile.css
+++ b/ui_components/src/profile/Profile.css
@@ -57,7 +57,7 @@
     height: 84px;
     position: relative;
 
-    font-size: 12px;
+    font-size: var(--font-min);
 
     display: inline-flex;
     flex-direction: column;
--- a/ui_components/src/profile/Profile.css
+++ b/ui_components/src/profile/Profile.css
@@ -1,0 +1,94 @@
+
+.profile_container {
+    max-width: 100%;
+
+    display: flex;
+    flex-direction: row;
+    flex-wrap: wrap;
+
+    padding: 20px;
+    box-sizing: border-box;
+}
+
+.profile_avatar_container {
+    width: 230px;
+    height: 250px;
+    margin-right: 20px;
+
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
+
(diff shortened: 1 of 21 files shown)


ui_components/src/sidebar/Sidebar.js
```diff
@@ -205,7 +121,13 @@ function Sidebar({ links, navigationAPIData, userAPIData }) {
...
     return (
         <div id="sidebar">
-            {displayCharacterSection(playerData, playerResources, playerSettings, regenTime, regenOffset)}
+            <SBCharacterProfile
+                playerData={playerData}
+                playerResources={playerResources}
+                playerSettings={playerSettings}
+                regenTime={regenTime}
+                regenOffset={regenOffset}
+            />
```
levimeahan, line 130: TODO: do this for the rest of the sections here

ui_components/src/sidebar/Sidebar.js
```diff
@@ -214,4 +136,111 @@ function Sidebar({ links, navigationAPIData, userAPIData }) {
...
+                            </svg>
+                        </div>
+                    </div>
+                </div>
+
+                <SBResourceBar
+                    resourceType="health"
+                    resourceAmount={playerResources.health}
+                    resourceMaxAmount={playerResources.max_health}
+                />
```
levimeahan, line 199: I love reusable components when they make sense

ui_components/src/profile/Profile.css
```diff
@@ -0,0 +1,94 @@
...
+.stat_list.attributes {
+    width: 50%;
+}
+
+.stat {
+    width: 100%;
+    height: 84px;
+    position: relative;
+
+    font-size: 12px;
```
aumu1031, line 60: Can we use `var(--font-size)` variables for this? Will make it easier if we want to adjust font size between layouts in the future or allow for global layout settings (e.g. dropdown -> font size: Large).

levimeahan: Yup I'll update that 👍

PR #347, opened 2023-06-28
0.9 Event Functionality

aumu1031:
Currently need to do manual DB update for mission_type = 6 (event) and mission rewards.

Reward column sample contents:
`[{"item_id":19,"chance":100,"quantity":3},{"item_id":20,"chance":50,"quantity":2}, {"item_id":21,"chance":5,"quantity":1}] `


classes/Mission.php
```diff
@@ -305,6 +308,46 @@ public static function start($player, $mission_id): Mission {
 
         $mission = new Mission($mission_id, $player);
 
+        // TEMP Event Logic
+        global $system;
+        if ($mission->mission_type == Mission::TYPE_EVENT) {
+            if ($mission->rank == 2) {
+                echo "here";
```
levimeahan, line 315: Is this debug/to be removed?

classes/Mission.php
```diff
@@ -305,6 +308,46 @@ public static function start($player, $mission_id): Mission {
...
+                $valid = false;
+                foreach ($system->event_data['easy'] as $location) {
+                    if ($player->location->x == $location['x'] && $player->location->y == $location['y']) {
+                        echo $player->location->x;
```
levimeahan, line 319: Is this debug/to be removed?

classes/Mission.php
```diff
@@ -320,4 +363,47 @@ public static function maxMissionRank(int $player_rank): int {
...
+    public static function processRewards(Mission $mission, User $player, System $system): string {
+        if (count($mission->rewards) > 0) {
+            try {
+                $reward_text = "";
+                // load player inventory
+                $player->getInventory();
+
+                // get items from DB
+                $reward_keys = [];
+                foreach ($mission->rewards as $item) {
```
levimeahan, line 376: If these are specifically item IDs, could we name this `$mission->reward_item_ids`?

classes/Mission.php
```diff
@@ -320,4 +363,47 @@ public static function maxMissionRank(int $player_rank): int {
...
+                $reward_key_string = '(' . implode(',', $reward_keys) . ')';
+                /** @var Item[] $reward_items */
+                $reward_items = array();
+                $result = $system->db->query("SELECT * FROM `items` where `item_id` IN {$reward_key_string}");
+                while ($row = $system->db->fetch($result)) {
+                    $reward_items[$row['item_id']] = Item::fromDb($row);
+                }
+                // roll RNG, add to inventory
+                foreach ($mission->rewards as $item) {
+                    if (rand(0, 100) <= $item['chance']) {
```
levimeahan, line 388:
```suggestion
                    if (mt_rand(0, 100) <= $item['chance']) {
```
Supposedly mt_rand is better than rand somehow. ALthough we could also use [random_int](https://www.php.net/manual/en/function.random-int.php)

classes/battle/BattleManager.php
```diff
@@ -729,6 +729,8 @@ protected function applyAttack(BattleAttack $attack, Fighter $user, Fighter $tar
 
         $text = $attack->jutsu->battle_text;
         $attack_jutsu_color = BattleManager::getJutsuTextColor($attack->jutsu->jutsu_type);
+        $has_element = ($attack->jutsu->element != "none");
```
levimeahan, line 732:
```suggestion
        $has_element = ($attack->jutsu->element != Jutsu::ELEMENT_NONE);
```

db/migrations/20230628144927_mission_reward_migration.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * Migrate Up.
+     */
+    public function up(): void
+    {
+        // Modify missions table
+        $this->execute("
+            ALTER TABLE `missions` 
+                ADD `rewards` VARCHAR(500) NOT NULL DEFAULT '[]';
```
levimeahan, line 16:
```suggestion
                ADD `rewards` VARCHAR(2000) NOT NULL DEFAULT '[]';
```
Lean really on the side of "way larger than we could ever fathom needing" with these if not using `TEXT`, it's tricky to track down when they're getting truncated

pages/event.php
```diff
@@ -165,6 +165,47 @@ function event() {
...
     }
 
+    if (isset($_GET['exchange'])) {
+        try {
+            // Change the item ID values based on dev/prod
+            // These are placeholder exchange options
+            switch ($_GET['exchange']) {
+                case "red_yen":
+                    $player->getInventory();
+                    if (!$player->hasItem("19")) {
```
levimeahan, line 175: Could we code the Item ID in event config somewhere? That will make it easier to swap for dev/prod

ui_components/src/home/Home.css
```diff
@@ -962,27 +988,73 @@ body {
...
+        max-height: 55px;
+    }
+    .home_lantern_5::after {
+        content: '';
+        box-shadow: 0 0 45px 3px #f9f8dd, 0 0 20px 2px #c78b7b;
+        position: absolute;
+        width: 10px;
+        height: 40px;
+        left: 5px;
+        top: 5px;
```
levimeahan, line 1088: This part is the same for each of these, could we pull it into a `.home_lantern::after` class?

ui_components/src/home/News.js
```diff
@@ -230,4 +234,11 @@ function NewsButtons({ loadNews, createPost, isAdmin }) {
             }
         </div>
     );
+}
+
+// Should probably be a global helper
+function clickOnEnter(event) {
+    if (event.key === "Enter") {
+        event.target.dispatchEvent(new MouseEvent('click', { bubbles: true }));
+    }
```
levimeahan, line 243: This is a global helper now, we can remove from here

levimeahan (approved):
Looks good! Excited for this event

classes/Mission.php
aumu1031: Good catch, I'll remove this.

classes/Mission.php
aumu1031: Same as above.

ui_components/src/home/Home.css
aumu1031: Resolved in commit.

ui_components/src/home/News.js
aumu1031: Resolved in commit.

classes/Mission.php
aumu1031: Resolved in commit.

PR #348, opened 2023-06-29
Home lantern fix, Event mission tweaks

aumu1031:


ui_components/src/profile/Profile.js
aumu1031, line 1: For clarity on `Profile.js` diff:
- Moved profile elements into rows 'profile_row_first' and 'profile_row_second'
- Wrapped `status_attributes` with `status_attributes_wrapper`
- Changes `status_info_section` from static widths to using equivalent flex-basis for scaling

pages/event.php
```diff
@@ -170,17 +172,54 @@ function event() {
...
                         throw new RuntimeException("You do not have enough of this item!");
                     }
-                    $player->items[$system->event_data['red_lantern_id']]->quantity -= 3;
-                    $player->addMoney("1000", "Event");
-                    $system->message("You have traded 3 Red Lanterns for 1000 yen!");
+                    $player->items[$system->event_data['red_lantern_id']]->quantity -= 1;
+                    if ($player->items[$system->event_data['red_lantern_id']]->quantity < 1) {
+                        unset($player->items[$system->event_data['red_lantern_id']]);
+                    }
+                    $player->addMoney("25", "Event");
```
levimeahan, line 187:
```suggestion
                    $player->addMoney(25, "Lantern Event");
```

levimeahan: Hmm actually would be nice if this all scaled off a single var like `$yen_per_lantern` so we don't have to update multiple numbers if we want to adjust the amount

pages/event.php
```diff
@@ -189,16 +228,228 @@ function event() {
...
+                    if ($player->items[$system->event_data['shadow_essence_id']]->quantity < 1) {
+                        unset($player->items[$system->event_data['shadow_essence_id']]);
+                    }
+
+                    $result = $system->db->query("SELECT * FROM `items` WHERE `item_id` = {$system->event_data['forbidden_jutsu_scroll']}");
+                    $player->items[$system->event_data['forbidden_jutsu_scroll']] = Item::fromDb($system->db->fetch($result), 1);
+
+                    $system->message("You have traded 1 Shadow Essence for a Forbidden Jutsu Scroll!");
+                    $player->updateInventory();
+                    $player->updateData();
```
levimeahan, line 452: There seems to be a lot of opportunity to extract common logic here into shared method(s) but I don't want to block this, we can do that later

templates/temp_event.php
```diff
@@ -9,17 +9,146 @@
...
+            </table>
+            <table class="table" style="margin-bottom: 25px">
+                <tr>
+                    <th>Cost</th>
+                    <th>Reward</th>
+                    <th>Exchange</th>
+                </tr>
+                <tr>
+                    <td>1x Red Lantern</td>
+                    <td>25 Yen</td>
```
levimeahan, line 42: These displays could also use the `yen_per_red_lantern` var mentioned above

ui_components/src/home/Home.js
```diff
@@ -168,10 +168,10 @@ function MainBannerSection({
                     <div className="title_version">{version}</div>
                 </div>
 
-                <div className={"home_lantern home_lantern_2"} style={{ zIndex: 0 }}><img src="/images/v2/decorations/lanternbig.png" /></div>
-                <div className={"home_lantern home_lantern_4"} style={{ zIndex: 1 }}><img src="/images/v2/decorations/lanternbig.png" /></div>
-                <div className={"home_lantern home_lantern_5"} style={{ zIndex: 1 }}><img src="/images/v2/decorations/lanternsmall.png" /></div>
-                <div className={"home_lantern home_lantern_7"} style={{ zIndex: 1 }}><img src="/images/v2/decorations/lanternsmall.png" /></div>
+                <div className={"home_lantern home_lantern_1"} style={{ zIndex: 1 }}><img src="/images/v2/decorations/lanternbig.png" /></div>
```
levimeahan, line 171: Hmm if z index is the same, why not set it on the `.home_lantern` class?

levimeahan (approved):

aumu1031:
To-do:
`rep_per_lantern` refactor
z-index to `home_lantern` class
Extract common event logic
Specify event in yen log

PR #351, opened 2023-06-30
Fix character creation, event tweaks/fixes

aumu1031:


pages/event.php
```diff
@@ -291,20 +317,20 @@ function event() {
                     if (!$player->hasItem($system->event_data['red_lantern_id'])) {
                         throw new RuntimeException("You do not have this item!");
                     }
-                    if ($player->items[$system->event_data['red_lantern_id']]->quantity < 500) {
+                    if ($player->items[$system->event_data['red_lantern_id']]->quantity < 100) {
```
levimeahan, line 320:
```suggestion
                    if ($player->items[$system->event_data['red_lantern_id']]->quantity < $shadow_essence_cost['red_lanterns']) {
```
We could centralize these costs a bit and make them easy to change

levimeahan (approved):
🚀 🌕

PR #353, opened 2023-07-01
Page fixes

aumu1031:


index.php
```diff
@@ -674,15 +676,14 @@
...
-    } else {
-        $layout = $system->fetchLayoutByName(System::DEFAULT_LAYOUT);
-    }
+
+    $layout = $system->fetchLayoutByName(System::DEFAULT_LAYOUT);
     $layout->renderBeforeContentHTML($system, $player ?? null, "Home", render_content: false, render_header: false, render_sidebar: false, render_topbar: false);
 
     // Display error messages
-    $system->printMessage();
+    if ($system->isDevEnvironment()) {
```
levimeahan, line 684: Huh, should this be based on layout? I think we'd always want to show these, it's just a matter of if the layout the user is on shows them a different way or not

levimeahan (approved):

index.php
aumu1031: System message is output to the form labels on home, this is for disabling the green text

PR #358, opened 2023-07-02
Fix notifications

aumu1031:


levimeahan (approved):

pages/event.php
```diff
@@ -227,6 +227,9 @@ function event() {
                     break;
                 case "red_rep":
                     $player->getInventory();
+                    if($player->calMaxRepGain(1) == 0) {
```
levimeahan, line 230:
```suggestion
                    if($player->weekly_rep >= $player->weekly_rep_cap) {
```
Hmm couldn't we also do this?

PR #365, opened 2023-07-04
Decouple avatar frame/style, lag fix (?), post-event logic, travel/card config

aumu1031:


aumu1031:
- Allow separate configuration for avatar frame and shape
- Fix achievement name for "Violet Lantern Holder"
- Fix gender option under register dropdown
- Handle case when event is unset
- Add Lantern Event mission/item IDs in static context
- Avoid row locking for APIs (allow for battle APIs)
- Add initial configuration for player cards
- Add travel animation/grid config
- Remove 10s delay from attacking

classes/achievements/_achievements.php
```diff
@@ -169,13 +169,13 @@
...
-        name: 'Purple Lantern Holder',
-        prompt: 'Obtain a Purple Lantern',
+        name: 'Violet Lantern Holder',
+        prompt: 'Obtain a Violet Lantern',
         rewards: [
             new AchievementReward(AchievementReward::TYPE_MONEY, 5_000)
         ],
         criteria_check_closure: function (System $system, User $player) {
-            return $player->hasItem($system->event->item_ids['violet_lantern_id']);
+            return $player->hasItem(LanternEvent::$static_item_ids['violet_lantern_id']);
```
levimeahan, line 178: Awesome! This is great

classes/event/LanternEvent.php
```diff
@@ -8,6 +8,27 @@ class LanternEvent extends Event {
     public array $item_ids = [];
     public array $mission_ids = [];
```
levimeahan, line 9: Can we remove these fields now that we have the static fields?

classes/notification/NotificationAPIManager.php
```diff
@@ -333,7 +333,7 @@ public function getUserNotifications(): array
             }
         }
         //Event
-        if ($this->system->event != null) {
+        if (isset($this->system->event)) {
```
levimeahan, line 336: Ah, we could alternatively fix this by defaulting `$event = null` in the System class. I suppose either is fine but IMO it's a little weird conceptually to ask if a class property is set (should always be set, just may or may not have a value)

classes/travel/TravelManager.php
```diff
@@ -332,87 +332,96 @@ public function fetchCurrentMapLocations(): array {
...
-                    map_id: 1,
-                    x: $event_mission['x'],
-                    y: $event_mission['y'],
-                    image: "/images/events/yokai_cropped.png",
-                    action_url: $this->system->router->getUrl("mission", [
-                        'start_mission' => $this->system->event->mission_ids['nightmare_mission_id'], 'mission_type' => 'event'
-                    ]),
-                    action_message: "Stop the Ritual",
-                );
+        if (isset($this->system->event)) {
```
levimeahan, line 335: This shouldn't be necessary, were you seeing an error here?
image

db/migrations/20230705160301_travel_settings_migration.php
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Migrate Up.
+     */
+    public function up(): void
+    {
+        // Modify user_settings table
+        $this->execute("
+            ALTER TABLE `user_settings`
+                ADD `travel_animation` VARCHAR(50) NOT NULL DEFAULT 'smooth',
+                ADD `travel_grid` VARCHAR(50) NOT NULL DEFAULT 'visible',
```
levimeahan, line 17: Personally - I would like to drop these two settings. I think we can fix the grid and pick one travel motion to align on

ui_components/src/home/RegisterForm.js
```diff
@@ -51,7 +51,7 @@ export function RegisterForm({ registerErrorText, registerPreFill }: RegisterFor
                         <div className="register_field_label">gender</div>
                         <select name="gender" defaultValue={registerPreFill.gender}>
                             <option value="Male">Male</option>
-                            <option value="Female">Male</option>
+                            <option value="Female">Female</option>
```
levimeahan, line 54: oof YIKES good catch

levimeahan (approved):

api/travel.php
```diff
@@ -3,7 +3,7 @@
 # Begin standard auth
 require_once __DIR__ . "/../classes.php";
 
-$system = API::init();
+$system = API::init(row_lock: false);
```
levimeahan, line 6:
```suggestion
$system = API::init(row_lock: true);
```
I think we need to enable locking for this one, high likelihood of causing conflicts

classes/event/LanternEvent.php
aumu1031: This is still accessed in object context as part of the Event class type.

classes/travel/TravelManager.php
aumu1031: Testing with/without event active, was throwing an error here I think?

PR #366, opened 2023-07-06
Display players on map

aumu1031:


api/battle.php
```diff
@@ -13,80 +13,159 @@
...
-
-$battle_result = $system->db->query("SELECT battle_type FROM battles WHERE `battle_id`='{$player->battle_id}' LIMIT 1");
-if($system->db->last_num_rows) {
-    $battle_data = $system->db->fetch($battle_result);
-
-    $battle_route = null;
-    foreach(Router::$routes as $page_id => $page) {
-        if(empty($page->battle_type)) {
-            continue;
+if (isset($_POST['request'])) {
```
levimeahan, line 16: Hmm, could we actually put this in the Travel API? I know technically this is a "battle" related action, but as is this API is pretty cleanly defined and only handles in-progress battles. The Travel API has more of an assorted collection of travel-related functionality related to movement and nearby players, where I think this would fit in as one other travel-related action

api/battle.php
```diff
@@ -13,80 +13,159 @@
...
+                // get user id off the attack link
+                $result = $system->db->query("SELECT `user_id` FROM `users` WHERE `attack_id`='{$attack_id}' LIMIT 1");
+                if ($system->db->last_num_rows == 0) {
+                    throw new RuntimeException("Invalid user!");
+                }
+
+                $attack_link = $system->db->fetch($result);
+                $attack_id = $attack_link['user_id'];
+
+                $user = User::loadFromId($system, $attack_id);
```
levimeahan, line 33:
```suggestion
                $target_user_id = $attack_link['user_id'];

                $user = User::loadFromId($system, $target_user_id);
```

api/battle.php
```diff
@@ -13,80 +13,159 @@
...
+                if ($user->last_death_ms > System::currentTimeMs() - (60 * 1000)) {
+                    throw new RuntimeException("Target has died within the last minute, please wait " .
+                        ceil((($user->last_death_ms + (60 * 1000)) - System::currentTimeMs()) / 1000) . " more seconds.");
+                }
+
+                if ($system->USE_NEW_BATTLES) {
+                    BattleV2::start($system, $player, $user, Battle::TYPE_FIGHT);
+                } else {
+                    Battle::start($system, $player, $user, Battle::TYPE_FIGHT);
+                }
```
levimeahan, line 78: We could put all this into a `TravelManager` method. 

The reason I think this particularly makes sense as a `Travel` type method is if we add functionality like variable attack distance, that's more related to the world and players moving around than it is the battle system itself. The battle system's concern starts and ends when a fight is kicked off (here on line 74-77)

style/new_geisha/new_geisha.css
```diff
@@ -489,3 +489,93 @@ h3 {
...
+.chat_user_name {
+    background: none !important;
+    clip-path: none !important;
+    cursor: pointer;
+    align-self: center;
+    --color-text-shadow: none !important;
+    flex-shrink: 0;
+}
+
+#chatReactContainer .chat_user_name.premiumUser:hover {
```
levimeahan, line 503: Ah, this is minor atm but I think we should move away from using the ReactContainer class names as that may change when we migrate the pages to an SPA framework. We should set a class name on the root div inside the relevant component (like Chat.js in this case) and attach these styles to that

ui_components/src/travel/Map.js
```diff
@@ -1,5 +1,5 @@
 
-export const Map = ({mapData}) => {
+export const Map = ({ mapData, scoutData, playerId }) => {
```
levimeahan, line 2: Could we add Flow types here to define the shape of these props? Can be a follow-up, I can help if needed

ui_components/src/travel/Map.js
```diff
@@ -186,11 +192,47 @@ function MapLocations({ locations, tileWidth, tileHeight }) {
...
+                        zIndex: 5,
                         }}></div>
                     }
                 </div>
             ))}
         </div>
     );
+}
+
+function ScoutLocations({ scoutData, tileWidth, tileHeight, playerId }) {
```
levimeahan, line 205: Wdyt about a more descriptive name, something like this?
```suggestion
function MapNearbyPlayers({ scoutData, tileWidth, tileHeight, playerId }) {
```

ui_components/src/travel/ScoutArea.js
```diff
@@ -64,7 +66,7 @@ const Player = ({
             </div>
             <div className='travel-scout-attack'>
                 {(player_data.attack === true && parseInt(player_data.battle_id, 10) === 0) && (
-                    <a href={attackLink + '&attack=' + player_data.attack_id}></a>
+                    <a onClick={() => attackLink(player_data.attack_id)}></a>
```
levimeahan, line 69: Huh, if this is now not a string with the attack link but is a function that conducts the attack, could we change the name to reflect that? Maybe
```suggestion
                    <a onClick={() => attackPlayer(player_data.attack_id)}></a>
```

aumu1031: Fixed in commit.

ui_components/src/travel/Map.js
aumu1031: Fixed in commit.

style/new_geisha/new_geisha.css
aumu1031: Moved back to chat CSS in commit, accidentally moved on copy/paste.

api/battle.php
aumu1031: Fixed in commit.

ui_components/src/travel/Map.js
aumu1031: Appreciated!

api/battle.php
aumu1031: Moved to travel.

api/battle.php
aumu1031: Moved to `TravelManager`.

api/travel.php
```diff
@@ -52,9 +52,14 @@
 
             $success = $TravelManager->updateFilter($filter, $filter_value);
             $TravelAPIResponse->response = TravelApiPresenter::travelActionResponse($success, $player, $TravelManager);
-
             break;
 
+        case 'AttackPlayer':
+            $target_user_id = $system->db->clean($_POST['target']);
+
+            $success = $TravelManager->attackPlayer($target_user_id);
```
levimeahan, line 60: This isn't actually the user's ID, this is the user's "attack ID" which is a random hash regenerated regularly to prevent bookmarking a specific user's attack link
```suggestion
            $target_attack_id = $system->db->clean($_POST['target_attack_id']);

            $success = $TravelManager->attackPlayer($target_attack_id);
```

classes/travel/TravelManager.php
```diff
@@ -488,6 +488,68 @@ public function fetchCurrentLocationPortal(): ?array {
...
+     */
+    public function attackPlayer($target_user_id): bool {
+        // get user id off the attack link
+        $result = $this->system->db->query("SELECT `user_id` FROM `users` WHERE `attack_id`='{$target_user_id}' LIMIT 1");
+        if ($this->system->db->last_num_rows == 0) {
+            throw new RuntimeException("Invalid user!");
+        }
+
+        $attack_link = $this->system->db->fetch($result);
+        $target_user_id = $attack_link['user_id'];
```
levimeahan, line 502:
```suggestion
    public function attackPlayer(string $target_attack_id): bool {
        // get user id off the attack link
        $result = $this->system->db->query("SELECT `user_id` FROM `users` WHERE `attack_id`='{$target_attack_id}' LIMIT 1");
        if ($this->system->db->last_num_rows == 0) {
            throw new RuntimeException("Invalid user!");
        }

        $target_user= $this->system->db->fetch($result);
        $target_user_id = $target_user['user_id'];
```

levimeahan (approved):

PR #367, opened 2023-07-07
Survival mission fix.

HitoriWoB:


levimeahan (approved):
wow is this all it took to fix survival-type missions?
:notlikethis:

PR #368, opened 2023-07-07
Reputation 2.3

HitoriWoB:


HitoriWoB:
Decay rates are placeholder numbers only currently. Just an assigned value to test the queries. Not sure how much effort we want to assign to each rank.

PR #370, opened 2023-07-08
Fix verification

aumu1031:
I'm confused as to why `!empty($_GET['act'])` is necessary for this to work for resend verification but `!empty($_POST['act'])` has apparently been working for the verification link, so this could use a second look and test on dev before deployment.


levimeahan (approved):
so weird. Yolo, let's test it!

PR #372, opened 2023-07-09
Fix chat colors, Restrict BL scout/stealth 1 square, 5m timer, Sort NearbyPlayers, Boost to %

aumu1031:


classes/battle/Fighter.php
```diff
@@ -319,12 +319,11 @@ public function calcDamage(Jutsu $attack, bool $disable_randomness = false): flo
...
 
         // Add non-BL damage boosts
-        $damage_boost = $off_boost - $off_nerf;
         if($this->system->debug['damage']) {
-            echo "Damage/boost/nerf/final boost: $damage / {$off_boost} / {$off_nerf} / {$damage_boost} <br />";
+            echo "Damage/boost/nerf: $damage / {$off_boost} / {$off_nerf}} <br />";
         }
 
-        $damage = round($damage + $damage_boost, 2);
+        $damage = round($damage * (1 + $off_boost) - $off_nerf, 2);
```
levimeahan, line 326: I know everybody probably knows PEMDAS but I always like to be explicit so it's clear it was intentional
```suggestion
        $damage = round(
          ($damage * (1 + $off_boost)) - $off_nerf, 
          2
        );
```

classes/travel/TravelManager.php
```diff
@@ -276,10 +287,35 @@ public function fetchNearbyPlayers(): array {
...
+                } else {
+                    // Sort by level if distances are equal
+                    if ($a->level > $b->level) {
+                        return -1; // $a comes before $b
+                    } elseif ($a->level < $b->level) {
+                        return 1; // $b comes before $a
+                    } else {
+                        return 0; // Objects are equal
+                    }
+                }
```
levimeahan, line 315: Huh, I wonder if this could lead to more misclicks. The more factors we introduce that can cause players to move up and down on the list, the easier it is to misclick and attack the wrong player.

I'm okay trying this out if you don't feel that's a concern, but suspect it might be better to keep the order as consistent as possible

ui_components/build/travel/Map.js
```diff
@@ -34,27 +35,27 @@ export const Map = ({
...
-  const map_offset_y = stage_offset_y * -1; // Calculate display values
+  const map_offset_y = stage_offset_y * -1;
 
+  // Calculate display values
   const map_width = parseInt(mapData.end_x) - parseInt(mapData.start_x) + 1;
   const map_height = parseInt(mapData.end_y) - parseInt(mapData.start_y) + 1;
   const PlayerStyle = {
     position: "absolute",
-    backgroundImage: "url(./" + mapData.player_icon + ")",
+    backgroundImage: `url(./${mapData.invulnerable ? 'images/ninja_head_grey.png' : mapData.player_icon})`,
```
levimeahan, line 58: We could just do `opacity: 0.5` instead of having a different image

ui_components/src/profile/Profile.js
```diff
@@ -219,12 +219,12 @@ function PlayerStats({ playerData, playerStats }) {
                 <div className="stat box-secondary">
                     <h3>Intelligence: {playerStats.intelligence.toLocaleString()}</h3>
                     <div className="badge">知能</div>
-                    <span className="ft-c3">lol</span>
+                    <span className="ft-c3"></span>
```
levimeahan, line 222: feels_bad_man

ui_components/src/profile/Profile.js
```diff
@@ -219,12 +219,12 @@ function PlayerStats({ playerData, playerStats }) {
...
                 </div>
                 <div className="stat box-secondary">
                     <h3>Willpower: {playerStats.willpower.toLocaleString()}</h3>
                     <div className="badge">根性</div>
-                    <span className="ft-c3">nope</span>
+                    <span className="ft-c3"></span>
```
levimeahan, line 227: why

ui_components/src/travel/Map.js
```diff
@@ -62,7 +62,7 @@ export const Map = ({ mapData, scoutData, playerId, ranksToView }) => {
 
     const PlayerStyle = {
         position: "absolute",
-        backgroundImage: "url(./" + mapData.player_icon + ")",
+        backgroundImage: `url(./${mapData.invulnerable ? 'images/ninja_head_grey.png' : mapData.player_icon})`,
```
levimeahan, line 65: same comment about `opacity: 0.5`

levimeahan (approved):

classes/travel/TravelManager.php
aumu1031: I think the main issue (now that there's a visual indicator on the map) is the attack link itself moving, so I'm confident this will be felt as an improvement. I'll keep an eye on player feedback and revert/adjust if necessary.

PR #374, opened 2023-07-13
Invulnerability logic, avatar options, site background, forbidden shop

aumu1031:


classes/forbidden_shop/ForbiddenShopAPIPresenter.php
```diff
@@ -0,0 +1,40 @@
...
+require_once __DIR__ . '/../event/LanternEvent.php';
+
+class ForbiddenShopApiPresenter {
+    /**
+     * @throws RuntimeException
+     */
+    public static function exchangeEventCurrencyResponse(ForbiddenShopManager $forbiddenShopManager, $event_name, $currency_name, $quantity)
+    {
+        return [
+            'message' => $forbiddenShopManager->exchangeEventCurrency($event_name, $currency_name, $quantity),
```
levimeahan, line 12: I think this call pattern is the opposite of how it should be - Any mutations should be called from manager => presenter, not the other way around. I can fix this in the follow up tweaks

classes/forbidden_shop/ForbiddenShopManager.php
```diff
@@ -0,0 +1,86 @@
...
+        switch ($event_name) {
+            case "festival_of_shadows":
+                $yen_per_lantern = LanternEvent::$static_config['yen_per_lantern'];
+                $red_lantern_id = LanternEvent::$static_item_ids['red_lantern_id'];
+                $blue_lantern_id = LanternEvent::$static_item_ids['blue_lantern_id'];
+                $violet_lantern_id = LanternEvent::$static_item_ids['violet_lantern_id'];
+                $gold_lantern_id = LanternEvent::$static_item_ids['gold_lantern_id'];
+                $shadow_essence_id = LanternEvent::$static_item_ids['shadow_essence_id'];
+                switch ($currency_name) {
+                    case "all":
```
levimeahan, line 32: Switch statement with one case doesn't make a lot of sense to me, we could rewrite as `if($currency_name == "all")` but what cases would that not be true in?

config/memes.php
```diff
@@ -674,24 +674,65 @@
...
+        'text' => ':smugbaka:',
+        'public' => true,
+    ],
+    [
+        'code' => ':cooljojo:',
+        'image' => "<img src='./images/memes/cooljojo.gif' />",
+        'text' => ':cooljojo:',
+        'public' => true,
+    ],
+];
```
levimeahan, line 738: We should actually delete this file, it doesn't matter anymore (memes are auto-generated from image files in the `memes` directory

levimeahan (approved):

PR #382, opened 2023-07-28
Reputation 2.5

HitoriWoB:


admin/reputation_simulator.php
```diff
@@ -0,0 +1,645 @@
...
+ * @var System $system
+ * @var User $user
+ */
+require_once __DIR__ . "/_authenticate_admin.php";
+$self_link = $system->router->base_url . 'admin/reputation_simulator.php';
+$data = null;
+$debug_data = null;
+
+class RepPlayerReputation extends UserReputation {
+    public function runDecay(RepPlayer $repPlayer, $sim_data) {
```
levimeahan, line 12:
```suggestion
    public function runDecay(RepPlayer $repPlayer, SimDataDto $sim_data) {
    
    // ...
    
    class SimDataDto {
       public function __construct(
          public float $decay_modifier,
          public int $daily_pvp_cap,
          public int $pvp_daily_con,
          // ...
       ) {}
    }
```
Totally optional here, but I think things are more readable if we get out of the habit of using associative arrays as objects and use typed data classes instead.

The syntax in this comment is PHP short-form class initialization to take out some of the boilerplate of writing it, see here for details: https://stitcher.io/blog/constructor-promotion-in-php-8

admin/reputation_simulator.php
```diff
@@ -0,0 +1,645 @@
...
+        // Update tier once more should rep loss have resulted in de-rank
+        $this->updateRepTier($sim_data);
+    }
+
+    public function updateRepTier($sim_data) {
+        $this->rank = self::tierByRepAmount($this->rep);
+        $REP_RANK = self::$VillageRep[$this->rank];
+        $this->rank_name = $REP_RANK['title'];
+        $this->weekly_cap = $REP_RANK['weekly_cap'];
+        $this->pvp_cap = ($sim_data['daily_pvp_cap']) ? floor($REP_RANK['weekly_pvp_cap'] * ($sim_data['pvp_daily_con']/100)) : $REP_RANK['weekly_pvp_cap'];
```
levimeahan, line 42: What is `$sim_data['pvp_daily_con']` ?

admin/reputation_simulator.php
```diff
@@ -0,0 +1,645 @@
...
+class RepPlayer {
+    const TYPE_CASUAL = 'casual';
+    const TYPE_AVERAGE = 'average';
+    const TYPE_ABOVE_AVERAGE = 'above_average';
+    const TYPE_NIGHTMARE = 'nightmare';
+    public static array $player_types = [
+        self::TYPE_CASUAL, self::TYPE_AVERAGE, self::TYPE_ABOVE_AVERAGE, self::TYPE_NIGHTMARE
+    ];
+
+    const WEEKLY_CAP_RATE = [
```
levimeahan, line 54: Just checking, is this % of cap we expect each type of player to hit in a week?

admin/reputation_simulator.php
```diff
@@ -0,0 +1,645 @@
...
+    'average_daily_easy' => 2,
+    'average_daily_med' => 5,
+    'average_daily_hard' => 9,
+    'decay_modifier' => UserReputation::DECAY_MODIFIER * 100,
+    'pvp_daily_con' => UserReputation::PVP_WEEKLY_CONVERSION * 100,
+    'reputation_data' => UserReputation::$VillageRep,
+    'run_player_types' => RepPlayer::$player_types,
+);
+
+function runRepSimulation(&$data, $sim_data, &$debug_data, $weeks = 4) {
```
levimeahan, line 136: Looking forward to playing with this! I think we should take this up one notch and instead of specifying each player by % of cap they will hit (which we have to calculate in spreadsheet) specify how many hours of special missions / missions / arena / etc we expect them to do, then calculate the rep from that. That way if we make tweaks in code to e.g. arena or mission gain or CD, we can live preview it without having to go to the spreadsheet and recalculate

classes/DailyTask.php
```diff
@@ -3,41 +3,61 @@
 class DailyTask {
     const NUM_PER_DAY = 3;
 
+    const SUB_TASK_EARN = 'earn';
     const SUB_TASK_WIN_FIGHT = 'win';
     const SUB_TASK_COMPLETE = 'complete';
+    const SUB_TASK_SKILL = 'skills';
+    const SUB_TASK_GEN = 'attributes';
```
levimeahan, line 10:
```suggestion
    const SUB_TASK_ATTRIBUTES = 'attributes';
```

classes/UserDailyTasks.php
```diff
@@ -0,0 +1,125 @@
+<?php
+class UserDailyTasks {
+    const TASK_RESET = 86400;
```
levimeahan, line 3:
```suggestion
    const TASK_RESET_COOLDOWN = 86400;
```
Could be a little more specific with naming here

classes/UserReputation.php
```diff
@@ -287,24 +341,34 @@ public function calcArenaReputation($player_level, $opponent_level) {
...
-        $this->encodePvpKills();
+        // Pvp reputation disabled
+        if(!self::PVP_REP_ENABLED) {
+            return 0;
+        }
 
-        if($winner == true) {
+        if($winner) {
+            // Set current kill
+            $this->recent_players_killed_ids_array[$opponent_user_id][] = time();
```
levimeahan, line 356: If this is doing mutations we should consider renaming, maybe instead of `calcPvPRep` it could be `handlePvPWin` and we handle adding/deducting in this method

pages/arena.php
```diff
@@ -248,10 +248,8 @@ function processArenaBattleEnd(BattleManager|BattleManagerV2 $battle, User $play
         $player->last_pvp_ms = System::currentTimeMs();
 
         // Daily Tasks
-        foreach ($player->daily_tasks as $task) {
-            if (!$task->complete && $task->activity == DailyTask::ACTIVITY_ARENA && $task->sub_task == DailyTask::SUB_TASK_WIN_FIGHT) {
-                $task->progress++;
-            }
+        if($player->daily_tasks->hasTaskType(DailyTask::ACTIVITY_ARENA)) {
+            $player->daily_tasks->progressTask(DailyTask::ACTIVITY_ARENA, 1);
```
levimeahan, line 252: Love this 🎉 so much cleaner

levimeahan (approved):
LFGGGGGG 🚢 🚀

classes/DailyTask.php
HitoriWoB: Resolved in #397

classes/DailyTask.php
```diff
@@ -215,21 +283,32 @@ public static function generateTask(User $user, string $task_activity, array $us
...
 
-        if($user->rank_num >= 3) {
-            $daily_tasks[] = DailyTask::generateTask($user, DailyTask::ACTIVITY_PVP);
+        // Rank 1 only
+        if($user_rank_num == 1) {
+            $daily_tasks[] = DailyTask::generateTask($user_rank_num, DailyTask::ACTIVITY_EARN_MONEY, $possible_difficulties);
+        }
+        // Rank 1 and 2 only - training tasks
+        if($user_rank_num <= 2) {
+            $daily_tasks[] = DailyTask::generateTask($user_rank_num, DailyTask::ACTIVITY_TRAINING, $possible_difficulties);
```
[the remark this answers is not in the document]
HitoriWoB: Resolved in #397

PR #384, opened 2023-08-02
Mission tweaks, chat fix, loss display, partial special gains

aumu1031:


classes/SpecialMission.php
```diff
@@ -389,32 +391,39 @@ public function nextEvent() {
...
+                    $reward_text .= ' You have gained ' . $rep_gain . " village reputation!";
+                }
             }
         }
 
         $stat_to_gain = $this->player->getTrainingStatForArena();
-        $stat_gain = self::$difficulties[$this->difficulty]['stats_per_mission']
-            + (max(0, $this->player->rank_num - 2) * 2);
+        $stat_gain = (int) (self::$difficulties[$this->difficulty]['stats_per_mission']
+            + (max(0, $this->player->rank_num - 2) * 2)) * $progress_modifier;
```
levimeahan, line 426: Could be a little more explicit about the intent here:
```suggestion
        $stat_gain = floor(
           (self::$difficulties[$this->difficulty]['stats_per_mission'] + (max(0, $this->player->rank_num - 2) * 2)
           * $progress_modifier
        );
```

classes/chat/ChatManager.php
```diff
@@ -227,7 +227,7 @@ private function fetchPosts(?int $starting_post_id = null, int $max_posts = self
     public function submitPost(string $message) {
         $chat_max_post_length = $this->maxPostLength();
 
-        $message_length = strlen(preg_replace('/[\\n\\r]+/', '', trim($message)));
+        $message_length = strlen(preg_replace('/(\\\\[rn])+/', '', trim($message)));
```
levimeahan, line 230: What use case is this targeting?

levimeahan (approved):

classes/chat/ChatManager.php
aumu1031: New lines are coming in as '\\r\\n'
The original capture group is missing all of them, we need to escape the '\' twice, and then capture 'r' or 'n' any number of times.

PR #386, opened 2023-08-03
Remove duplicate stage display, Ramen at colosseum, Bloodline boost as float

aumu1031:


classes/Route.php
```diff
@@ -4,6 +4,7 @@ class Route {
     const NOT_IN_VILLAGE = 0;
     const IN_VILLAGE_OKAY = 1;
     const ONLY_IN_VILLAGE = 2;
+    const VILLAGE_OR_ARENA = 3;
```
levimeahan, line 7:
```suggestion
    const VILLAGE_OR_COLOSSEUM = 3;
```
Not to be confused with the normal arena

classes/Router.php
```diff
@@ -170,6 +170,23 @@ public function assertRouteIsValid(Route $route, User $player): void {
 
                 throw new RuntimeException("You must be in your village to access this page!");
             }
+
+            if ($route->village_ok === Route::VILLAGE_OR_ARENA && !$player->location->equals($player->village_location)) {
+                // hardcoding location, #dontlookplz
+                $arena_coords = new TravelCoords(16, 15, 1);
```
levimeahan, line 176: hmm... Okay but let's at least hardcode this in one place like System or TravelManager and reference it from there here and in healing page

levimeahan (approved):
Could we squeeze in a themed overhaul to make this not the Ramen Shop when it's at the stadium? So different location name + different food names (maybe stick-themed or handheld foods like teriyaki skewers, dango (sweet dumpling on stick, anko's favorite), onigiri, etc

PR #393, opened 2023-08-08
Undo Clamp, Mission Rep, New Layout (dev), Battle Stage Logic, Remove Survival

aumu1031:


layout/sumu.php
```diff
@@ -0,0 +1,164 @@
+<?php
+
+function getSumuLayout(System $system, bool $enable_mobile_layout): Layout {
```
levimeahan, line 3: Is the HTML structure for this different from New Geisha? I think we should not be duplicating all this if possible - Could we re-use New Geisha's HTML but pass different CSS links in?

levimeahan (approved):

PR #396, opened 2023-08-10
PvE Rep for UC Spars, Rep tweaks, Partial special mission gains, Spar link at UC, Chat row locking, Mission fix

aumu1031:


aumu1031:
Win: 5 Rep
Draw: 3 Rep
Lose: 2 Rep

Has requirements:
Must be at UC
Must be within 5 levels
Battle must be 3+ turns
Uses mission/arena rep cooldown

aumu1031:
image
- Daily tasks use minimum mission rank.
- Increase baseline daily task Rep (mission/arena) from 15->30, set PvP/training task to half mission/arena.

aumu1031:
* Double special mission Rep, 1/2/3/4 -> 2/4/6/8. Allows player to break even on Rep/time assuming always ends in PvP loss.

classes/UserDailyTasks.php
HitoriWoB, line 1: Great idea! No sense in forcing lower-tiered content if it's avoidable. Maybe I should have it roll mid-tiered missions, then higher can be completed if desired?

classes/UserReputation.php
```diff
@@ -118,33 +118,33 @@ class UserReputation {
...
             DailyTask::ACTIVITY_EARN_MONEY => 3,
-            DailyTask::ACTIVITY_ARENA => 20,
-            DailyTask::ACTIVITY_TRAINING => 8,
-            DailyTask::ACTIVITY_MISSIONS => 20,
-            DailyTask::ACTIVITY_PVP => 15,
+            DailyTask::ACTIVITY_ARENA => 40,
+            DailyTask::ACTIVITY_TRAINING => 20,
+            DailyTask::ACTIVITY_MISSIONS => 40,
+            DailyTask::ACTIVITY_PVP => 20,
         ],
```
HitoriWoB, line 148: I feel like grind/reward is pretty balanced here. Should we pump up money tasks to 5, 10, 15? It's strictly rank 1 content so not very abuseable.

HitoriWoB (commented):
Think reputation changes look good!

classes/UserReputation.php
aumu1031: 5/10/15 sounds good! Will make the change.

classes/UserDailyTasks.php
aumu1031: Not sure how they're currently rolled but I like the variety.
Most players are going to grind the A-Rank outside of dailies so having C/D/B tasks breaks up the monotony.

levimeahan (approved):

PR #397, opened 2023-08-10
DailyTask update & Rep Benefit shifting

HitoriWoB:
~~Daily tasks now allow for training tasks for Ranks Akademi-sai and Genin. This task will default to jutsu training and open up skills and attributes once at least 1 point has been gained in each (currently excludes intelligence & willpower stats).~~

PvP daily task has now been changed to roll either strictly pvp completes or pvp completes & spars. The chance of getting a strict pvp task is 20+{weekly_pvp_rep}% chance. This means once 80 PvP rep has been gained, player battle tasks will always be strictly pvps.

Reputation benefits have been reworked:
Tier 3: 5% consumable shop discount & more efficient extended trainings (down from T6)
Tier 4: 5% gear shop discount & more efficient long trainings (down from T6)
Tier 5: 5% jutsu scroll shop discount
Tier 6: Partial training gains (down from T7)

We have also been discussing as a content team on the requirements for seats. Clan seats will be somewhere between tiers 3 - 5 and village seats will be somewhere between tiers 5 - 7.


db/migrations/20230811005835_yen_gain_update.php
```diff
@@ -0,0 +1,121 @@
...
+            ADD COLUMN `last_malicious_ip` TEXT AFTER `login_attempt_time`
+        ");
+        // Chat deletion
+        $this->execute("ALTER TABLE `chat` ADD COLUMN `deleted` INT(1) NOT NULL DEFAULT 0");
+        // NPC stuff
+        $this->execute("ALTER TABLE `ai_opponents` CHANGE `money` `money_multiplier` SMALLINT(3) NOT NULL;");
+        // Rank 1 AI
+        $this->execute("UPDATE `ai_opponents` SET `money_multiplier`=1 WHERE `ai_id`=1");
+        $this->execute("UPDATE `ai_opponents` SET `money_multiplier`=2 WHERE `ai_id`=2");
+        $this->execute("UPDATE `ai_opponents` SET `money_multiplier`=2 WHERE `ai_id`=3");
```
levimeahan, line 33: Hmm, what's the intent of this? This strikes me as something we could statically calculate

classes/Currency.php
```diff
@@ -0,0 +1,128 @@
+<?php
+class Currency {
```
levimeahan, line 2: This is cool! I almost wonder if we should take this a step further and make it a `UserCurrency` class that has fully named variables for every type of user currency, calls would look like:

`$player->money->amount` => `$player->currency->money`
`$player->money->add()` => `$player->currency->addMoney()`

I'm not sure which way is better, okay with your approach as is but something to ponder

classes/DailyTask.php
```diff
@@ -54,10 +60,11 @@ class DailyTask {
         DailyTask::ACTIVITY_MISSIONS => 'Missions',
         DailyTask::ACTIVITY_TRAINING => [
             DailyTask::SUB_TASK_SKILL => 'points',
-            DailyTask::SUB_TASK_GEN => 'points',
+            DailyTask::SUB_TASK_ATTRIBUTES => 'points',
             DailyTask::SUB_TASK_JUTSU => 'levels',
         ],
```
levimeahan, line 65: To make this usable, we could take off the sub tasks and make it a flat stat-based task - train any skill or attribute by XYZ points. Then we don't assign the task if player is already at total stat cap

classes/DailyTask.php
```diff
@@ -243,7 +264,8 @@ public static function generateTask(int $user_rank_num, string $task_activity, s
             $rep_reward_mod += UserReputation::DAILY_TASK_PVP_WIN_MOD;
         }
 
-        $difficulty_multiplier = 1;
+        $base_multiplier = mt_rand($user_rank_num, $user_rank_num + 2);
```
levimeahan, line 267: We could move these base multipliers for each type of activity to Currency class as well so it's easy to see/change them all in one place and compare "specials gives 10x, arena gives 2x, special takes 50x as long as arena, maybe we give it more" etc

levimeahan:
I think you have a bunch of commits from https://github.com/elementum-games/shinobi-chronicles/pull/396 in this, can you try rebasing on latest main?

HitoriWoB:
> I think you have a bunch of commits from https://github.com/elementum-games/shinobi-chronicles/pull/396 in this, can you try rebasing on latest main?

I just the merge. There'd abumdant commits from working inside github @ work

db/migrations/20230811005835_yen_gain_update.php
HitoriWoB: This is intended to take the place of the static money gains that are currently set for each AI. Example: Instead of saying Academy Bully gives 210 yen, it gives 1 multiplier.

We could totally remove this and allow arena logic and remove set rates in the NPC DB.

classes/DailyTask.php
HitoriWoB: I like this as a change. I'm going to keep the jutsu training as a task set to break up monotony of the task-type and provide an alternative if skills are capped on when the task reroll happens.

classes/DailyTask.php
HitoriWoB: I was thinking similar

levimeahan (approved):

classes/news/NewsAPIPresenter.php
```diff
@@ -10,7 +10,12 @@ class NewsApiPresenter {
     public static function newsPostResponse(NewsManager $newsManager, System $system, int $num_posts = 8): array {
         return array_map(
             function (NewsPostDto $post) use ($system) {
-                $message = $system->parseMarkdown(stripslashes($post->message), true);
+                if(!System::LOCAL_HOST) {
+                    $message = $system->parseMarkdown(stripslashes($post->message), true);
+                }
+                else {
+                    $message = nl2br(stripslashes($post->message));
+                }
```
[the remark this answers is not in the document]
HitoriWoB: The parseamrkdown breaks my local version and I've confirmed all resources are up-to-date.

index.php
```diff
@@ -113,22 +118,27 @@
...
-                    $system->db->query(
-                        "UPDATE `users` SET `failed_logins`= 0 WHERE `user_id`='{$result['user_id']}' LIMIT 1"
-                    );
-                }
 
                 $player = User::loadFromId($system, $_SESSION['user_id']);
                 $player->loadData();
                 $player->last_login = time();
+                $player->failed_logins = 0;
+                $player->last_malicious_ip = null;
```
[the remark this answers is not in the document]
HitoriWoB: I think clearing it should be fine. We log all malicious attempts into the DB. This particular user member is only used in determining lockout cd

classes/DailyTask.php
HitoriWoB: Will complete this in next PR

classes/Currency.php
HitoriWoB: I like the `$player->currency->addMoney()` methodology

HitoriWoB:
This is ready for final review/mrege

HitoriWoB:
Changes here have been upgraded and are out of date. See #406 for current data.

PR #399, opened 2023-08-17
Faction Missions

aumu1031:
- New mission type `FACTION`
- Allow setting Origin for a mission (in place of village)
- Allow setting Travel stage to a specific location
- Added currency item use type
- Renamed "Event" purchase type to "Reward"


aumu1031:
Adjusted Speed offense ratio from 0.25 => 0.35
Calculates to a 10% increase in damage output for a 50k BL /  50k Offense / 150k Speed build
Calculates to a 4% increase in damage output for a 85k BL / 85k Offense / 80k Speed build

aumu1031:
Made header coordinates update on movement
Made regen rates visible on sidebar hover

api/forbidden_shop.php
```diff
@@ -45,6 +48,17 @@
             $message = $ForbiddenShopManager->buyForbiddenJutsu($jutsu_id);
             $player->getInventory();
 
+            $response->data = [
+                'message' => html_entity_decode($message, ENT_QUOTES),
+                'playerInventory' => UserApiPresenter::playerInventoryResponse($player)
+            ];
+            break;
+        case "exchangeFavor":
+            $item_id = $system->db->clean($_POST['item_id']);
```
levimeahan, line 57: If this is a numeric ID, you could do `(int)$_POST['item_id']` instead

admin/entity_constraints.php
```diff
@@ -517,12 +518,18 @@
...
         'data_type' => 'int',
         'input_type' => 'radio',
-        'options' => [1 => 'Village', 2 => 'Clan', 3 => 'Team', 4 => 'Special', 6 => 'Event'],
+        'options' => [1 => 'Village', 2 => 'Clan', 3 => 'Team', 4 => 'Special', 6 => 'Event', 7 => 'Faction'],
     ],
     'money' => [
         'data_type' => 'int',
         'input_type' => 'text',
     ],
+    'origin' => [
```
levimeahan, line 527:
```suggestion
    'custom_start_location' => [
```
see below

classes/Mission.php
```diff
@@ -43,6 +47,7 @@ class Mission {
     public $stages;
     public $money;
     public $rewards = [];
+    public ?TravelCoords $origin;
```
levimeahan, line 50:
```suggestion
    public ?TravelCoords $custom_start_location;
```
I was a little unsure reading this originally, I think a more specific name like this would help make it easier to understand how this field is used

classes/battle/Fighter.php
```diff
@@ -10,7 +10,7 @@ abstract class Fighter {
     const BLOODLINE_OFFENSE_RATIO = self::SKILL_OFFENSE_RATIO * 0.8;
     const BLOODLINE_DEFENSE_MULTIPLIER = 35;
 
-    const SPEED_OFFENSE_RATIO = 0.25;
+    const SPEED_OFFENSE_RATIO = 0.35;
```
levimeahan, line 13: What's the rationale on this one?

classes/forbidden_shop/ForbiddenShopAPIPresenter.php
```diff
@@ -26,6 +27,19 @@ public static function eventDataResponse()
...
+    /**
+     * @throws RuntimeException
+     */
+    public static function exchangeDataResponse()
+    {
+        return [
+            'exchangeData' => [
+                'ayakashiFavor' => ForbiddenShopManager::AYAKASHI_FAVOR,
+                'favorExchange' => ForbiddenShopManager::FAVOR_EXCHANGE,
+                'factionMissions' => ForbiddenShopManager::FACTION_MISSIONS
```
levimeahan, line 39: Huh, a little confused why we need to return static variables in an API response. How will these be used in the frontend - Could we pass them in as props during component initialization?

classes/forbidden_shop/ForbiddenShopManager.php
```diff
@@ -11,6 +11,19 @@ public function __construct(System $system, User $player) {
         $this->player = $player;
     }
 
+    const AYAKASHI_FAVOR = 131;
```
levimeahan, line 14:
```suggestion
    const AYAKASHI_FAVOR_ITEM_ID = 131;
```

classes/forbidden_shop/ForbiddenShopManager.php
```diff
@@ -11,6 +11,19 @@ public function __construct(System $system, User $player) {
...
+
+    const FAVOR_EXCHANGE = [
+        127 => 1000, // forbidden scroll
```
levimeahan, line 17:
```suggestion
        LanternEvent::$static_item_ids['forbidden_jutsu_scroll_id'] => 1000, // forbidden scroll
```

levimeahan: This might not work, maybe we need to change the lantern event id array to a constant. But I think we should try to keep it as a single source of truth for what ID each item is

db/migrations/20230817004323_forbidden_mission_migration.php
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Migrate Down.
+     */
+    public function down(): void
+    {
+        // Modify missions table
+        $this->execute("
+            ALTER TABLE `missions`
+                DROP `origin`;
+        ");
```
levimeahan, line 38: Need to delete the new missions here as well

classes/battle/Fighter.php
aumu1031: Looking at the balance of high builds relative to what they're supposed to counter and what counters them.

The most common two types of builds are 0 speed and ~85k speed. These two types of builds are roughly balanced versus each other - the same bloodline at 0 and 85k speed will deal almost identical damage to each other. 

A build with speed beyond 85k should expect to gain advantage versus 85k while becoming weaker against builds closer to 0 speed. The issue is the tradeoff, where speed beyond 85k is only marginally helpful versus other builds while being substantially weaker versus 0. 

So you have a case where high speed (like >110k) is an automatic loss to 0 speed without having a notable advantage versus ~85k. 

The proposed change is very small in terms of overall power but I think is enough to make speed worthwhile:
```Adjusted Speed offense ratio from 0.25 => 0.35
Calculates to a 10.2% increase in damage output for a 50k BL / 50k Offense / 150k Speed build
Calculates to a 3.9% increase in damage output for a 85k BL / 85k Offense / 80k Speed build```

classes/forbidden_shop/ForbiddenShopAPIPresenter.php
aumu1031: Good point, API isn't necessary here.

db/migrations/20230817004323_forbidden_mission_migration.php
aumu1031: Fixed in commit.

classes/forbidden_shop/ForbiddenShopManager.php
aumu1031: Fixed in commit.

classes/forbidden_shop/ForbiddenShopAPIPresenter.php
aumu1031: Fixed in commit.

classes/Mission.php
aumu1031: Fixed in commit.

admin/entity_constraints.php
aumu1031: Fixed in commit.

api/forbidden_shop.php
aumu1031: Fixed in commit.

levimeahan (approved):

classes/battle/Fighter.php
levimeahan: Makes sense, let's do it 👍

PR #405, opened 2023-08-18
Map Update, Misc Fixes

aumu1031:


db/migrations/20230817004323_forbidden_mission_migration.php
```diff
@@ -61,7 +61,7 @@ public function down(): void
         // Delete from items table
         $this->execute("
             DELETE FROM `items`
-                WHERE `item_id` = 131;
+                WHERE `item_id` = 132;
```
HitoriWoB, line 64: I think this was supposed to stay 131? Otherwise we're removing the favor item.

aumu1031: This is the down() migration so it should be yeah?

pages/settings.php
aumu1031, line 1: Messy diff, changes begin at line 231

classes/Village.php
```diff
@@ -8,13 +8,19 @@ class Village {
...
+    // to-do: we should restructure how village data is being saved
+    // player village should reference the village ID and this constructor should get row by ID
     public function __construct($system, $village) {
         $this->system = $system;
 
         $this->name = $village;
         $this->kage_name = $this->getKageName();
         $this->coords = $this->setVillageCoords();
+        $this->region_id = $this->getRegionId($this->system, $this->name);
+        $this->village_id = $this->getVillageId($this->system, $this->name);
```
levimeahan, line 23: Not the biggest deal, but where possible let's do things in a single DB query. Could we query for both values here and set them in the constructor?

classes/travel/Region.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public static function coordInRegion(RegionCoords $coord, $polygon) {
+        $x = $coord->x;
+        $y = $coord->y-1;
+        $numVertices = count($polygon);
+        $inside = false;
+
+        for ($i = 0, $j = $numVertices - 1; $i < $numVertices; $j = $i++) {
+            $xi = $polygon[$i][0];
```
levimeahan, line 67:
```suggestion
            $xi = $polygon[$i][RegionCoords::X];
```

classes/travel/Region.php
```diff
@@ -0,0 +1,82 @@
...
+        return $new_region;
+    }
+
+    public static function coordInRegion(RegionCoords $coord, $polygon) {
+        $x = $coord->x;
+        $y = $coord->y-1;
+        $numVertices = count($polygon);
+        $inside = false;
+
+        for ($i = 0, $j = $numVertices - 1; $i < $numVertices; $j = $i++) {
```
levimeahan, line 66: Could you write a comment about how this check is intended to work?

ui_components/src/sidebar/Sidebar.js
```diff
@@ -76,11 +76,15 @@ function Sidebar({ links, navigationAPIData, userAPIData }) {
         }
     };
 
+    const parentClick = (e) => {
+        e.target.parentElement.click();
+    }
+
     // Content
-    function displaySection(section_data, title, toggleNav) {
+    function displaySection(section_data, title, toggleNav, sidebarCollapse, parentClick) {
```
levimeahan, line 84: This should really be pulled out into a component at this point

ui_components/src/sidebar/Sidebar.js
```diff
@@ -137,10 +141,10 @@ function Sidebar({ links, navigationAPIData, userAPIData }) {
                 regenTime={regenTime}
                 regenOffset={regenOffset}
             />
-            {displaySection(userMenu, "Player Menu", toggleNav)}
-            {displaySection(activityMenu, "Action Menu", toggleNav)}
-            {displaySection(villageMenu, "Village Menu", toggleNav)}
-            {staffMenu.length ? displaySection(staffMenu, "Staff Menu", toggleNav) : null}
+            {displaySection(userMenu, "Player Menu", toggleNav, playerSettings.sidebar_collapse, parentClick)}
```
levimeahan, line 144: When the function is pulled out to a component, these can be

```js
<DisplaySection
   sectionData={userMenu}
   title="Player Menu"
   toggleNav={toggleNav}
   sidebarCollapsed={playerSettings.sideBar_collapse}
   parentClick={parentClick}
/>
```

levimeahan (approved):

classes/travel/Region.php
aumu1031: The math of it is a little complicated, it's a ray-cast algorithm I found that's fairly lightweight. Added a comment.

classes/travel/Region.php
aumu1031: By this you mean associate 0 to X and 1 to Y explicitly since they're coordinate pairs inside the array?

ui_components/src/sidebar/Sidebar.js
aumu1031: I agree, need to take a pass at the original React work I did using functions instead of functional components.
Will do this as a follow-up.

classes/Village.php
aumu1031: Fixed.

classes/Mission.php
```diff
@@ -185,14 +185,48 @@ public function nextStage($stage_id): int {
...
+                else if (isset($this->current_stage['target_type']) && $this->current_stage['target_type'] != "default") {
+                    // get village in home region
+                    if ($this->current_stage['target_type'] == "home_village") {
+                        $village_query = $this->system->db->query("SELECT `x`, `y` FROM `region_locations`
+                            INNER JOIN `regions` ON `regions`.`region_id` = `region_locations`.`region_id`
+                            WHERE `regions`.`region_id` = '{$this->player->village->region_id}' AND `type` = 'village'
+                        ");
+                        $village_result = $this->system->db->fetch_all($village_query);
+                        $random_village = array_rand($village_result);
+                        $location = $this->rollLocation(new TravelCoords($village_result[$random_village]['x'], $village_result[$random_village]['y'], 1));
```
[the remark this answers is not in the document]
aumu1031: Fixed.

PR #407, opened 2023-08-20
Filter Bloodline by Jutsu Type

MiteyDrill:
ezgif com-video-to-gif

Should be possible to filter Bloodlines by their Jutsu Type. 
___

If a bloodline has a jutsu type with 'fire', then bloodlines with jutsu element of 'fire' will show. 
Filtering is done by adding a class name to the <td> html elements during the php construction foreach loop, and then using the <select onchange> event to filter out classes that have the jutsu element types. 

- One issue is <td> elements can have more than one class of the same name like <td classname='fire fire none'>

### DIFF
--- a/templates/bloodlineList.php
+++ b/templates/bloodlineList.php
@@ -29,23 +29,50 @@
         }
     }
 
-    //filter by category
+    //used to keep track of current filters
+    let typeFilter = 'none';
+    let elementFilter = 'none';
+
+    function filterByElement(type){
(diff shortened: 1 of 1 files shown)


MiteyDrill:
image
Added a way to filter via Jutsu Type (nin/gen/tai) - the method of filtering is a little clumsy but I think it works...

levimeahan (approved):
Looks good! Let's do it

PR #409, opened 2023-08-25
Training Page -> React Component

MiteyDrill:
I wanted to mess with the CSS or add some engaging visuals to the Training Page since its a little barebone and lonely even tho its visited quite often by users, but I didn't want to add even more to it if its still PHP based. I thought maybe some interactable or visual ui components to liven things up would be nice, but I thought if I made any visual changes to the page it would make life harder for if or when the page would be switched over to a React based component. 

I took some time to put down a skeleton of Training Page built with React Components as a concept. The components are simple and its not 1:1 visually or pretty just yet but I wanted to make a proof of concept and get a review before I put in more time.

The React Components on the top of the Training Page should work 1:1 with the PHP equivalent under it.

Let me know if this React code is any helpful. 

Oh also, the first or second commit says I changed like 20+ files but those are the React Build files I may have accidentally rebuilt when trying to run the react watch-ui script, but I think nothing about them really changed.. myb...

image
image


pages/training_v2.php
aumu1031, line 1: Something you can do here is check if the current layout is V2 (`New Geisha`) and split the display based on that.
See the Profile page as an example of this.
This way we can tie the `training_v2` display specifically to the newer layouts and use standardized CSS variables. It also means you don't have to worry about backwards compatibility with older layouts.

MiteyDrill: I'll look into how profile2 is displayed in the New Geisha layout and see if I can replicate that

MiteyDrill:
image
I added some cleaner CSS

MiteyDrill:
@aumu1031 I did some testing, I checked each rank, I checked when player has team benefits, i'm not sure how the partial gains thing works but I set it up to reflect what the old partial gains looked like. I checked jutsu and bloodline, and no jutsu no bloodline, etc. I think its somewhat working as intended o7

levimeahan (approved):
Looks pretty solid! Let's make this dev-only for now until the API is ready and visuals are cleaned up but after that I'm okay to merge. Sorry for turbo-long delay on review 🙇‍♂️

MiteyDrill:
No worries, I've been busy with school so I haven't had much time to work on it anyways its all good

aumu1031:
Can we hold off on pushing to dev for now?
Main thing is if we need to fix/tweak anything, that's extra dev time for upcoming release.

HitoriWoB:
@MiteyDrill, any chance we could get the devEnv wrap as suggested by Arth? I have some bugs I need to work out of the training code and associating my changes with your react changes would be the ideal scenario.

MiteyDrill:
I'm not sure what devEnv is @HitoriWoB

HitoriWoB:
In the training page it would be:

`
if($system->isDevEnviornment()) {
//Your new training code
} else {
//original training code
}
`

MiteyDrill:
@HitoriWoB oh okay I understand now, yeah I'll make that change and update the commit later this evening

HitoriWoB:
Thank you, I appreciate it :) I'm not worried about the training gain changes/fixes, I'll review them all with my fixes

PR #414, opened 2023-08-27
Map Update Bug Fixes

aumu1031:


aumu1031:
- Fixed special missions moving onto village tiles
- Adjusted mission rep rewards for increased times (missions will be reworked accordingly)

ui_components/build/travel/Travel.js
```diff
@@ -544,9 +544,7 @@ function TravelActions({
     onPointerUp: handlePointerUp,
     onPointerLeave: handlePointerUp,
     onContextMenu: e => {
-      if (e.nativeEvent.pointerType === "touch") {
-        e.preventDefault();
-      }
+      e.preventDefault();
```
levimeahan, line 547: Ah, why was this needed?

levimeahan (approved):
LGTM!

ui_components/build/travel/Travel.js
aumu1031: People were reporting that when holding down movement keys on mobile the context menu is popping up after a few seconds. For some reason the pointerType === "touch" check isn't inclusive enough so I'm trying the nuclear option.

PR #415, opened 2023-08-29
Stats chart graph

MiteyDrill:
image

A radial Chart that displays a range from [0 - 5] of skill power depending on the max skill.

Pros:
It's a little cool.

Cons:
Only on the [New Geisha] layout at the moment.
No Bloodline Skill.
Takes up Achievement row space.
Depending on character skill builds, certain shapes of the graph might be wonky like lets say someone maxes out Taijutsu and Intelligence and nothing else then instead of a fill area between each line element such as what can be seen above between Cast Speed and Ninjutsu, there would just be straight lines for Intelligence and Taijutsu, such as what is seen above in the image for Genjutsu.


MiteyDrill:
Wait I screwed up

PR #416, opened 2023-08-29
Stats chart graph 2

MiteyDrill:
image

A radial Chart that displays a range from [0 - 5] of skill power depending on the max skill.

Pros:

- It's a little cool.

Cons:

- Only on the [New Geisha] layout at the moment.

- No Bloodline Skill.

- Takes up Achievement row space.

- Depending on character skill builds, certain shapes of the graph might be wonky like lets say someone maxes out Taijutsu and Intelligence and nothing else then instead of a fill area between each line element such as what can be seen above between Cast Speed and Ninjutsu, there would just be straight lines for Intelligence and Taijutsu, such as what is seen above in the image for Genjutsu.


aumu1031:
Recommending that this replace the stat section as a toggle:
image
And then toggle control positioned above the stat bar at top right:
image

MiteyDrill:
image
image

Its a little cumbersome at the moment but it works

MiteyDrill:
image

MiteyDrill:
image
Charts now show up on legacy layouts

aumu1031:
I recommend placing Cast Speed between Ninjutsu and Genjutsu so you can have a consistent shape,
Also ideally placing Bloodline dynamically.
- If Genjutsu Boost: place between Cast Speed and Genjutsu
- If Ninjutsu Boost: place between Cast Speed and Ninjutsu
- If Taijutsu Boost: place between Speed and Taijutsu

MiteyDrill:
I'll try to get to that sometime this weekend o7

levimeahan (approved):

PR #422, opened 2023-09-04
Initial War Mechanics, Travel Overhaul, Sumu

aumu1031:
- Refactor TravelManager/API/Travel.js: decoupled `map_locations`, `region_objectives` and `map_objectives`
- Refactor `Village` constructor (by ID, has support for original by string)
- Add `NearbyPatrols` data and display logic (`patrols` + `caravans`) - icons placeholder
- Add `village_resources`
- Add `village_relations`
- Add `operations` <-- internal name for Infiltrate / Raid / Reinforce war actions
- Restyle Travel page and add `QuickScout` button (tracks player at top of list, can click to move/attack)
- Added `ReactTransitionGroup` reference for travel animations
- Enabled light-theme 'Sumu'
- Add styling for old pages on new layouts
- Added support for Double EXP Event
- Misc: increased post-battle delay slightly so that it isn't dependent on lag
- Misc: added daily user count
- Misc: redirect player from travel -> battle if in PvP (prevent using travel page to scout while in battle)
- Misc: redirect player from all pages -> battle if PvP complete (prevent sitting in a completed battle for PvP immunity)
- Misc: Daze battle text fix
- Misc: change chat focus to input box on meme/quote
- Misc: reverted refactor of notification API, now triggers player update
- Misc: fixed a bug with new training page that allowed starting a train at stat cap
- Misc: added requested memes
- Misc: added jutsu names / battle text colors
- Misc: add turn count to battle
- Misc: fixed decimal value on sidebar regen
- Misc: added random BL purchase option, disabled gaining Lesser
- Misc: fixed version display home page
- Misc: center-aligned usernames in tables
- Misc: updated hand seals
- Misc: updated Bloodline stats section for clarity
- Misc: removed decimals from display
- Misc: fixed mission bug with search radius
- Misc: auto scroll to details on jutsu page click
- Misc: Heal now works even on first turn when at full HP
- Misc: AI no longer use standard strike unless they have no other jutsu
- Misc: Auto-refresh battle (uses same logic as `BattleManager` for turn complete)
- Misc: No longer removed from bloodline clan on village change


classes/travel/NearbyPatrol.php
aumu1031, line 1: Note: Display and pathfinding functional, other attributes like battles are WIP

aumu1031:
Sample Patrol table data:
image
Sample Caravan table data:
image

aumu1031:
Sample Travel updated:
image

pages/settings.php
aumu1031, line 1: Enabling Sumu on live, also decoupling display name of layouts from the value in dropdown.
image

ui_components/build/travel/Map.js
aumu1031, line 1: Adding NearbyPatrols section
Decoupling map locations, map objectives, and region objectives
Using ReactTransitionGroup for animations

ui_components/build/travel/ScoutArea.js
aumu1031, line 1: Added a QuickScout section
- Mirrors the scouting tracker for the first player in the list
- Can be clicked to move in that direction
- Can be clicked to trigger an attack on the player
Should make PvP on mobile significantly more accessible

aumu1031:
Updated bloodline purchase:
image
Updated battle view:
image

aumu1031:
Styling old content for New Geisha and Sumu:
image
image

api/notification.php
aumu1031, line 1: Reverted change that allowed previewing training updates for notification display.
Now row locks and performs the full user update.

api/travel.php
aumu1031, line 1: - Add NearbyPatrol data (displays patrols and caravans on the map)
- Add `travel_message` -> allows passing messages from travel manager to the front-end display
- Add `BeingOperation` and `CancelOperation` -> for Infiltrate/Reinforce/Raid actions

api/user.php
aumu1031, line 1: `response_data['battle_url']` is used to redirect the user to battle when a battle finishes
this prevents the exploit of sitting in a completed battle for pvp immunity

classes/Mission.php
aumu1031, line 1: Fixes bug that caused mission stages with a count (e.g. search X times) to change the center radius
So a search would get progressively farther from the intended starting point

classes/System.php
aumu1031, line 1: Adding war_enabled flag to disable functionality on live

classes/User.php
aumu1031, line 1: Added `operation` data for user, is set to the ID of the user's active operation (like a special mission).

Also reverted earlier change to user update that allowed notifications to preview player update.

classes/battle/Battle.php
aumu1031, line 1: Added new battle type to handle AI battles for war mechanics.

classes/battle/BattleEffectsManager.php
aumu1031, line 1: Removed decimals from values and added CSS class targets for colored text styling.

classes/battle/BattleManager.php
aumu1031, line 1: Removed decimals from values and added CSS class targets for colored text styling.

classes/notification/NotificationAPIManager.php
aumu1031, line 1: Restored previously removed player update in notification API.

classes/travel/Travel.php
aumu1031, line 1: Increasing post-battle delay slightly so latency doesn't prevent being attacked

classes/travel/TravelApiPresenter.php
aumu1031, line 1: Added new data for front end
`$travelManager->checkOperation($player->operation);` <-- this checks the interval to progress a player's active operation

classes/village/VillageRelation.php
aumu1031, line 1: Every `village` should be initialized with an array `village_relations` containing 4 instances of `VillageRelation`

classes/war/Operation.php
aumu1031, line 1: - Has constants for operation definitions
- `HandleCompletion` and `HandleFailure` are WIP pending final requirements
User experience of operations: enter tile of village or castle, click start operation, progress bar ticks (player locked into position), drains pools every interval (WIP), then affects the target location on completion. 
Reinforce - increases target Defense and generates 1 resource of the location's type
Infiltrate - decreases target Defense and steals 1 resource of the location's type
Raid - decreases target Defense, deals damage, and steals 1 resource of the location's type

classes/war/WarManager.php
aumu1031, line 1: Contains functionality relating to war mechanics

pages/bloodline.php
aumu1031, line 1: Visual changes to make bloodline mechanics more clear to users (addresses common causes of confusion)

pages/members.php
aumu1031, line 1: Making members view consistent, adding a daily user count

pages/villageHQ.php
aumu1031, line 1: Style, centering the first column for site consistency

pages/war.php
aumu1031, line 1: Placeholder page to display war AI battles
Can also be used later as an entry point to display other content relating to war (?)

templates/battle/action_prompt.php
aumu1031, line 1: Using hand seal tooltips to display numbers (players getting confused over which seal is what # internally)
Adding classes to seals for targeted CSS

templates/battle/battle_interface.php
aumu1031, line 1: Add turn number, remove decimals

templates/bloodlineList.php
aumu1031, line 1: Remove Lesser bloodlines from display list

templates/jutsu_page.php
aumu1031, line 1: Styles, and making the page auto-scroll to jutsu details on click

templates/premium/premium.php
aumu1031, line 1: Updating bloodline purchase display

templates/settings.php
aumu1031, line 1: Changing layout display to a custom string (e.g. Shadow Ribbon [Legacy])

ui_components/src/travel/ScoutArea.js
aumu1031, line 1: Added a QuickScout section

    Sits at the center of the directional buttons
    Mirrors the scouting tracker for the first player in the list
    Can be clicked to move in that direction
    Can be clicked to trigger an attack on the player
    Should make PvP on mobile significantly more accessible

ui_components/src/travel/Map.js
aumu1031, line 1: Adding NearbyPatrols section
Decoupling map locations, map objectives, and region objectives
Using ReactTransitionGroup for animations
Adding display for operations, village HP and defense, village icons

ui_components/src/travel/Travel.js
aumu1031, line 1: Updating the overall display layout and styles

classes/NPC.php
aumu1031, line 1: Removed "standard strike" from the AI move pool unless AI has no other jutsu option

db/migrations/20230904154152_initial_war_migration.php
```diff
@@ -0,0 +1,198 @@
...
+        $this->execute("ALTER TABLE `region_locations` ADD `resource_penalty` INT(11) DEFAULT 0");
+        $this->execute("ALTER TABLE `region_locations` ADD `defense` INT(11) NOT NULL DEFAULT '0' AFTER `resource_penalty`");
+        $this->execute("UPDATE `region_locations` SET `resource_production` = 25");
+        $this->execute("UPDATE `region_locations` SET `defense` = 75 where `type` = 'castle'");
+        $this->execute("UPDATE `region_locations` SET `defense` = 50 where `type` = 'village'");
+        $this->execute("UPDATE `region_locations` SET `name` = 'Village' where `type` = 'village'");
+        $this->execute("UPDATE `region_locations` SET `name` = 'Castle' where `type` = 'castle'");
+
+        // Update villages table - add column for resources
+        $this->execute("ALTER TABLE `villages` ADD `resources` VARCHAR(200) DEFAULT '[]'");
```
levimeahan, line 38:
```suggestion
        $this->execute("ALTER TABLE `villages` ADD `resources` VARCHAR(1000) DEFAULT '[]'");
```
What's the longest conceivable length of serialized resources you could see being in this column?

pages/bloodline.php
```diff
@@ -93,31 +93,31 @@ function bloodline() {
 			'ratio_multiplier' => 100
 		),
 		'ninjutsu_boost' => array(
-			'text' => "[BL_SKILL] * [RATIO] -> <span class='amount'>[AMOUNT] extra ninjutsu offense (+[PERCENTAGE]%)</span>"
+			'text' => "[BL_SKILL] * [RATIO] -> <span class='amount'>[AMOUNT] extra Ninjutsu offense</span><br><span style='padding-left: 15px; font-size: smaller; font-style: italic'>Boost decreases to 75% strength as Bloodline Skill exceeds Ninjutsu skill</span>"
```
levimeahan, line 96: Nice-to-have in a later follow up would be an actual value here shown, e.g. if player's BL skill is greater than Nin skill we add a `(Power currently reduced by 17% due to Bloodline skill exceeding Ninjutsu skill)` message, but for players whose BL skill is less than offense, we show nothing

pages/members.php
```diff
@@ -261,7 +261,9 @@ function members(): void {
                     "SELECT COUNT(`user_id`) as `online_users` FROM `users` WHERE `last_active` > UNIX_TIMESTAMP() - $online_seconds"
                 );
                 $online_users = $system->db->fetch($online_users_result)['online_users'];
-                $list_name = 'Online Users (' . $online_users . ' currently online)';
+                $daily_users_result = $system->db->query("SELECT COUNT(`user_id`) as `daily_users` FROM `users` WHERE `last_active` > UNIX_TIMESTAMP() - $daily_seconds");
+                $daily_users = $system->db->fetch($daily_users_result)['daily_users'];
+                $list_name = 'Online Users (' . $online_users . ' online, ' . $daily_users . ' daily)';
```
levimeahan, line 266:
```suggestion
                $list_name = 'Online Users (' . $online_users . ' online, ' . $daily_users . ' today)';
```

ui_components/src/travel/Map.js
```diff
@@ -41,8 +41,8 @@ export const Map = ({ mapData, scoutData, playerId, ranksToView, strategicView,
        How do we calculate the starting coordinate in this example? We need to offset the first visible tile by +2 which
        is equal to player X - stage midpoint X.
      */
-    const stage_offset_x = player_x - stage_midpoint_x;
-    const stage_offset_y = player_y - stage_midpoint_y;
+    const stage_offset_x = player_x - stage_midpoint_x - 1;
+    const stage_offset_y = player_y - stage_midpoint_y - 1;
```
levimeahan, line 45: Huh, why is this needed?

ui_components/src/travel/Map.js
```diff
@@ -123,14 +123,92 @@ export const Map = ({ mapData, scoutData, playerId, ranksToView, strategicView,
...
+                                        fill="none"
+                                        strokeDasharray="100"
+                                        strokeDashoffset={100 - ((100 / 100) * mapData.operation_interval)}
+                                        transform="rotate(-90, 24.5, 24)"
+                                    />
+                                </svg>
+                            </div>
+                            </>
+                        }
+                    </div>
```
levimeahan, line 211: This should get pulled out into its own component, there's already so much going on in this one

ui_components/src/travel/Map.js
```diff
@@ -205,102 +283,257 @@ function MapGridLines({ tileWidth, tileHeight, stageOffsetX, stageOffsetY, stage
...
+            </ReactTransitionGroup.TransitionGroup>
+        </div>
+    );
+}
+
+function MapObjectives({ objectives, tileWidth, tileHeight }) {
+    return (
+        <div className='map_objectives'>
+            {objectives.map((objective) => (
+                <div key={objective.id}
```
levimeahan, line 322:
```suggestion
                <div key={`objective:${objective.id}`}
```
I think this way you can have a totally unique ID that avoids the "two elements in different locations have the same numerical ID" trap. I'm not really sure why React pushes for this, but they seem to suggest having truly unique keys?

ui_components/src/travel/Map.js
```diff
@@ -205,102 +283,257 @@ function MapGridLines({ tileWidth, tileHeight, stageOffsetX, stageOffsetY, stage
...
+    return (
+        <div id="patrol_locations" className='map_locations'>
+            {patrolData
+                .map((patrol, index) => (
+                    <div key={patrol.patrol_id + '_' + patrol.patrol_type}
+                        className='map_location'
+                        style={{
+                            cursor: "pointer",
+                            transform: `translate3d(${(patrol.target_x - 1) * tileWidth}px, ${(patrol.target_y - 1) * tileHeight}px, 0)`,
+                            backfaceVisibility: "hidden",
```
levimeahan, line 459: Interestingly I didn't see any performance difference with this on/off. What do you see on your machine?

ui_components/src/travel/Travel.js
```diff
@@ -702,4 +797,109 @@ function directionFromKeysPressed(directionKeysPressed) {
...
+                        <feFuncA type="linear" slope="1" />
+                    </feComponentTransfer>
+                    <feComposite in="floodColor" in2="opacityAdjustedBlur" operator="in" result="coloredBlur" />
+                    <feMerge>
+                        <feMergeNode in="coloredBlur" />
+                        <feMergeNode in="SourceGraphic" />
+                    </feMerge>
+                </filter>
+            </defs>
+        </svg>
```
levimeahan, line 901: mind blown

levimeahan (approved):
Holy moly. Awesome job! No real blocking comments, although a lot of stuff I didn't read the code deeply enough to see how it will work and we'll just have to test on dev

ui_components/src/travel/Map.js
aumu1031: I was having performance issues with hardware acceleration turned off originally, so this is probably only helpful for slow machines.

ui_components/src/travel/Map.js
aumu1031: The player icon is incorrectly centered on live, this shifts it so its directly in the middle

ui_components/src/travel/Map.js
aumu1031: Completed.

classes/travel/Patrol.php
```diff
@@ -0,0 +1,238 @@
...
+                $village_result = $system->db->query("SELECT `x`, `y` FROM `maps_locations`
+                    INNER JOIN `villages` ON `villages`.`map_location_id` = `maps_locations`.`location_id`
+                    WHERE `village_id` = {$this->village_id} LIMIT 1");
+                $village_result = $system->db->fetch($village_result);
+                // get castle location
+                $castle_result = $system->db->query("SELECT `x`, `y` FROM `region_locations`
+                    WHERE `region_id` = {$this->region_id} AND `type` = 'castle' LIMIT 1");
+                $castle_result = $system->db->fetch($castle_result);
+                switch ($this->caravan_type) {
+                    case "resource":
```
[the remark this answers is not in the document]
aumu1031: Completed.

classes/travel/Patrol.php
```diff
@@ -0,0 +1,238 @@
...
+                        $points[] = ['x' => $point['x'], 'y' => $point['y']];
+                    }
+                }
+                foreach ($result as $point) {
+                    if ($point['type'] == 'village') {
+                        $points[] = ['x' => $point['x'], 'y' => $point['y']];
+                    }
+                }
+                break;
+            case "caravan":
```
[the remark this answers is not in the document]
aumu1031: Completed.

classes/travel/Patrol.php
```diff
@@ -0,0 +1,238 @@
...
+        foreach ($row as $key => $value) {
+            $this->$key = $value;
+        }
+        $this->patrol_type = $patrol_type;
+    }
+    public function setLocation(System $system) {
+        $points = [];
+        $loop = false;
+        switch ($this->patrol_type) {
+            case "patrol":
```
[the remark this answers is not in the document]
aumu1031: Completed.

classes/battle/BattleEffectsManager.php
```diff
@@ -556,14 +554,14 @@ public function hasDisplays(Fighter $fighter): bool {
     }
 
     public function getDisplayText(Fighter $fighter): string {
-        return $this->system->db->clean(
+        return htmlspecialchars_decode($this->system->db->clean(
```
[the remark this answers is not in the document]
aumu1031: Noted. This change was to allow battle effect text to sit in the database correctly as HTML markup - I wonder if there's a better way to handle that? Strangely only the battle effect text (and not the battle damage, descriptions, etc.) was being encoded.

classes/travel/Patrol.php
```diff
@@ -0,0 +1,238 @@
...
+                    if ($point['type'] == 'village') {
+                        $points[] = ['x' => $point['x'], 'y' => $point['y']];
+                    }
+                }
+                break;
+            case "caravan":
+                $loop = false;
+                // get village location
+                $village_result = $system->db->query("SELECT `x`, `y` FROM `maps_locations`
+                    INNER JOIN `villages` ON `villages`.`map_location_id` = `maps_locations`.`location_id`
```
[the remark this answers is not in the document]
aumu1031: Marking as to-do before Kage/War release, taking a pass at every related table and adding indexes as appropriate.
In the short term it probably won't have an impact but some of these tables will grow large.

ui_components/src/travel/Map.js
aumu1031: Hmmm... these **should** be totally unique since the ID comes from the database primary key, 
Or is this more of a safety net we should apply to keys in general?

classes/travel/Travel.php
```diff
@@ -5,8 +5,8 @@
 class Travel {
 
     const TRAVEL_DELAY_MOVEMENT = 300; // milliseconds
-    const TRAVEL_DELAY_PVP = 500; // milliseconds
-    const TRAVEL_DELAY_AI = 250; // milliseconds
+    const TRAVEL_DELAY_PVP = 1500; // milliseconds
+    const TRAVEL_DELAY_AI = 750; // milliseconds
```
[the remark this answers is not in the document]
aumu1031: You can sit on someone who is in battle and repeatedly clicking the attack icon, but even then its still possible for someone to finish their battle and walk away due to latency. I have low ping to the server, so for people further away this is an even bigger issue.

I think 1.5s is long enough to make sure there's a clickable attack window without making travel feel restrictive.

classes/travel/TravelManager.php
```diff
@@ -403,24 +435,266 @@ public function fetchCurrentMapLocations(): array {
...
+                    image: "/images/v2/icons/reticle.png",
+                    objective_type: "target",
                 );
             }
             if ($this->user->mission_stage['action_type'] == 'search') {
                 $mission_result = $this->system->db->query("SELECT `name` FROM `missions` WHERE `mission_id` = '{$this->user->mission_id}' LIMIT 1");
                 $mission_location = TravelCoords::fromDbString($this->user->mission_stage['last_location']);
-                $objectives[] = new MapObjectiveLocation(
+                $objectives[] = new MapObjective(
+                    id: null,
```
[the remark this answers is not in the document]
aumu1031: Good catch!
MapObjectives are basically the way to pass any additional content to display on the map. So before that was mission objective, events, and regions. Since the IDs aren't necessarily coming from a DB primary key we can start the keys at specific values.

The player's mission target can always be set to 1 for example, event IDs can always start at 1000, etc. That's essentially how they worked before. Just to make sure we can have consistent IDs for React to update without risk of overlapping.

classes/war/WarManager.php
```diff
@@ -0,0 +1,184 @@
...
+                    Operation::OPERATION_RAID => System::unSlug(Operation::OPERATION_TYPE[Operation::OPERATION_RAID]),
+                ];
+                break;
+            case 'default':
+                return $valid_operations;
+        }
+        return $valid_operations;
+    }
+
+    function checkBeginPatrolBattle(Patrol $patrol) {
```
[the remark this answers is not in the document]
aumu1031: Completed

config/memes.php
```diff
@@ -735,4 +735,34 @@
...
+        'image' => "<img src='./images/memes/sadgies.png' />",
+        'text' => ':sadgies:',
+        'public' => true,
+    ],
+    [
+        'code' => ':applecat:',
+        'image' => "<img src='./images/memes/applecat.gif' />",
+        'text' => ':applecat:',
+        'public' => true,
+    ],
```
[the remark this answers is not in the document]
aumu1031: O

cron_jobs/caravan_cron.php
```diff
@@ -0,0 +1,131 @@
+<?php
+session_start();
+
+require_once __DIR__ . '/../classes/System.php';
+require_once __DIR__ . '/../classes/Village.php';
+require_once __DIR__ . '/../classes/User.php';
+
+const BASE_RESOURCE_PRODUCTION = 25;
+const BASE_CARAVAN_TIME_MS = 300000;
```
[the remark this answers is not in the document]
aumu1031: Agreed! Completed

cron_jobs/caravan_cron.php
```diff
@@ -0,0 +1,131 @@
...
+
+    // get resources from caravans
+    $caravan_result = $system->db->query("SELECT * FROM `caravans`");
+    $caravan_result = $system->db->fetch_all($caravan_result);
+    foreach ($caravan_result as $caravan) {
+        $caravan_resources = json_decode($caravan['resources'], true);
+        // add resource to village
+        foreach ($caravan_resources as $resource_id => $quantity) {
+            $villages[$caravan['village_id']]->addResource($resource_id, $quantity);
+        }
```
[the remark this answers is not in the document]
aumu1031: That's correct, since it runs every hour it assumes all caravans have now reached their destination.
If we need to rework this at some point, we can use the stop time of the caravan (they have a set start and duration so determining end time is how we know when to stop displaying them).

db/migrations/20230904154152_initial_war_migration.php
aumu1031: 8 resources with IDs, plus their value and wrapper
{"1":9999999,"2":9999999,"3":9999999,"4":9999999,"5":9999999,"6":9999999,"7":9999999,"8":9999999} <- 97 characters

pages/bloodline.php
aumu1031: I like it! Taking note for when we update the bloodline page, it's one of the higher priority to replace IMO.

pages/members.php
aumu1031: Completed

classes/travel/TravelManager.php
```diff
@@ -403,24 +435,266 @@ public function fetchCurrentMapLocations(): array {
...
+                  AND `from_id`={$this->user->location->map_id}
+                  AND `active`=1
+                  "
+        );
+
+        if ($this->system->db->last_num_rows) {
+            $portal_data = $this->system->db->fetch($result);
+        }
+
+        return $portal_data;
```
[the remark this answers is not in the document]
aumu1031: Admittedly I've not touched portals but I'll add it as a task, maybe a good first issue?

classes/travel/TravelApiPresenter.php
```diff
@@ -40,6 +44,13 @@ public static function mapDataResponse(User $player, TravelManager $travelManage
             'region_coords'     => $travelManager->getCoordsByRegion($regions),
             'spar_link'         => $system->router->getUrl('spar'),
             'colosseum_coords'  => $travelManager->getColosseumCoords(),
+            'region_objectives' => $travelManager->fetchRegionObjectives(),
```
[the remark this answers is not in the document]
aumu1031: Completed

classes/travel/Patrol.php
```diff
@@ -0,0 +1,238 @@
...
+    public string $patrol_type;
+    public string $alignment;
+    const DESTINATION_BUFFER_MS = 5000; // duration non-looped patrols should appear at their destination
+    public function __construct(array $row, string $patrol_type) {
+        foreach ($row as $key => $value) {
+            $this->$key = $value;
+        }
+        $this->patrol_type = $patrol_type;
+    }
+    public function setLocation(System $system) {
```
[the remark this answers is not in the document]
aumu1031: I think `setLocation` makes more sense here since it only sets these values in the end:
```
$this->current_x = $position['x'];
$this->current_y = $position['y'];
```

classes/travel/Patrol.php
```diff
@@ -0,0 +1,238 @@
...
+        $points = [];
```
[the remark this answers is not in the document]
aumu1031: Agreed, completed

classes/travel/Patrol.php
```diff
@@ -0,0 +1,238 @@
...
+                // if adjacent to given point consider at that point
+                if (abs($x - $points[$i]['x']) + abs($y - $points[$i]['y']) < 2) {
+                    return ['x' => $points[$i]['x'], 'y' => $points[$i]['y']];
+                }
+
+                return ['x' => $x, 'y' => $y];
+            }
+        }
+
+        return ['x' => $points[0]['x'], 'y' => $points[0]['y']];
```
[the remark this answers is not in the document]
aumu1031: That final line is a fail-safe if the current location is not found along the path, I think I added this as a check to make sure patrols start at their initial location correctly.

classes/travel/Patrol.php
```diff
@@ -0,0 +1,238 @@
+<?php
+
```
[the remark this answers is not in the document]
aumu1031: Added a bunch of comments to the code. Basic overview:

- Patrols have a set of `route_locations`. They move from the first location to the next in series. There are two differences between Patrol and Caravan types:
 1.) Travel speed OR Travel time can be set. 
 Patrol-types move at a given speed, and some Patrols have a longer path (time) to travel than others as a result.
 Caravans have a set travel time, so all Caravans will reach the Village at the same time (if they spawn together).
 2.) They can be set to loop, or not loop. Patrol-types will loop forever while Caravans stop at their final destination.

The actual input we need:
* $t: The current time.
* $T_start: The start time of the cycle.
* $T_cycle: The total duration of one cycle.
* $points: An array of route_locations.
* $loop: Flag for whether the path should loop back to the start or not.

$T_cycle is unknown for Patrols (they use a set speed instead) so we calculate it first using:
totalIntermediatePoints() * travel_interval
So for example, 20 squares at 500ms speed would be 10 seconds.

The actual calculate of the patrol's current position uses `calculatePositionNormalized` which is a heavy edit of `calculatePosition` (a basic piecewise linear interpolation formula). The main functional distinctions are:
 - We can set a boolean for looping (which ignores the last segment on the path if true).
 - The original formula has a tendency to skip the route_location points (the castles/villages) for some regions so this forcefully makes the patrols enter the villages/castles longer than at other points.

PR #430, opened 2023-09-19
v0.9.2 Bugs/Tweaks

aumu1031:
- Enable Ayakashi's Abyss during events
- Battle timer for spectate now refreshes correctly
- Battle timer during pre-heal now counts down correctly
- Fixed style for invulnerable players
- Fixed style for chat buttons
- PvE Rep CD from 60s -> 30s
- Rep from Spars now works in village as well (same-village players running to colosseum is counterintuitive)
- Moved delete button away from quote button (mods accidentally press)
- Fix rounding of bloodline boost on bloodline page
- Notification close on click
- Host `react-transition-group.js` directly (fix site loading slow)

### DIFF
--- a/classes/battle/BattleEffectsManager.php
+++ b/classes/battle/BattleEffectsManager.php
@@ -359,7 +359,7 @@ class BattleEffectsManager {
             $attack_jutsu_color = BattleManager::getJutsuTextColor($effect->damage_type);
 
             if($residual_damage_resisted > 0) {
-                $this->addDisplay($target, $target->getName() . " takes " . "<span class=\"battle_text_{$effect->damage_type}\" style=\"color:{$attack_jutsu_color}\">" . round($damage) . "</span>" . " residual damage (resists " . round($residual_damage_resisted) . " residual damage)");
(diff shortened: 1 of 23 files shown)


classes/battle/BattleEffectsManager.php
aumu1031, line 1: Adding colored text for the resist portion of residual damage, already does for direct damage

classes/travel/TravelManager.php
aumu1031, line 1: Was originally hidden during events because we planned to use forbidden jutsu scrolls as a general currency and wanted to restrict players earning one jutsu per event.

ui_components/build/topbar/Topbar.js
aumu1031, line 1: `closeNotification()` can optionally redirect after closing notification

ui_components/build/topbar/TopbarNotification.js
aumu1031, line 1: Clicking a notification that can be closed should redirect + close rather than requiring multiple clicks

ui_components/src/chat/Chat_new.js
aumu1031, line 1: Mods were hitting close by accident when trying to quote

levimeahan (approved):

PR #432, opened 2023-09-24
Style Tweaks

aumu1031:
- Minimum travel button width
- Travel button stack vertically
- Rounding on bloodline page
- Jutsu dropdown size
- New Geisha button style


levimeahan (approved):
LGTM 🚢

PR #434, opened 2023-09-26
War Phase 2, Reputation Update

aumu1031:
- Completed operations (Infiltrate/Reinforce/Raid/Loot)
- Completed loot system (gain from Infiltrate/Loot actions, claimed at village, stolen via PvP)
- Completed region capture system (Raid on Castle changes control and updates related data)
- Completed patrol system (chance to spawn across 3 tiers, on defeat sets respawn timer)
- Completed caravan system (collects resources from region, moves from castle -> village over 5 minutes, can be looted)
- Completed region system (village/castles produce resources/hour, have regen/hour, have defense value that shifts by 1/hour toward baseline, villages give a share of their regen to castle based on current health remaining)

Operation Cost: 150/interval, 5 intervals => 750 stamina/chakra
Operation Time: 12s/interval, 5 intervals => 60s

Infiltrate:
Interval: take 1 random resource
Completion: decrease defense by 1

Reinforce:
Interval: restore level/2 health (simple value scaled to player level, can rework as more mechanics are added)
Completion: increase defense by 1

Raid:
Interval: deal (level - defense) damage  (simple value scaled to player level, can rework as more mechanics are added)
Completion: decrease defense by 1
Special: on interval, if health == 0 takes region control, set castle health to 25%, update patrols/caravans

Loot:
Interval: steal 2 random resources
Special: only has 1 interval, so takes 12s per Loot action

https://docs.google.com/spreadsheets/d/1zfbyLluH-DxqOWM64fmWR3j_MggmaK9PdVbEyNBl7SU/edit#gid=1566094071

Reputation
image
image

Other
- Fixed bug with quick action button on travel tracking allies
- Rework spar rep gain so it grants a proportion of 10 (7/3 win, 5/5 draw, 3/7 loss) from (5/2, 3/3, 2/5)
- Removed rep gain CD (only restricted arenas and D-rank which have built-in time limiters)
- Made free stat transfer equal to 1% of stat total as baseline (2500 Jonin, 500 Chuunin). Added +50% free stat transfer as Forbidden Seal and T5 Rep benefit (max 2% of stat total daily)
- Added jutsu training speed bonus as Rep T4 reward (25% faster)


cron_jobs/caravan_cron.php
```diff
@@ -22,30 +23,36 @@
     if ($player->staff_manager->isHeadAdmin()) {
         if (isset($_GET['run_script']) && $_GET['run_script'] == 'true') {
             if ($system->environment === System::ENVIRONMENT_PROD) {
+                $debug = false;
+                if (isset($_GET['debug'])) {
+                    $debug = $system->db->clean($_GET['debug']);
```
levimeahan, line 28: We could call this `dry_run` to be a little more indicative of what it does, but this is ok 👍

cron_jobs/region_cron.php
```diff
@@ -0,0 +1,186 @@
...
+        $run_ok = true;
+    }
+
+    if ($run_ok) {
+        hourlyRegion($system);
+        $system->log('cron', 'Hourly Region', "Regions have been processed.");
+    } else {
+        $system->log('cron', 'Invalid access', "Attempted access by " . $_SERVER['REMOTE_ADDR']);
+    }
+}
```
levimeahan, line 73: Hmm, starting to feel like we should pull this into some kind of standard cron auth function

pages/battle.php
```diff
@@ -171,12 +177,18 @@ function processBattleFightEnd(BattleManager|BattleManagerV2 $battle, User $play
...
+        if ($player->system->db->last_num_rows > 0) {
+            $result .= "You have claimed the loot being carried by your opponent.[br]";
+        }
 
     } else if ($battle->isOpponentWinner()) {
         $result .= "You lose. You were taken back to your village by some allied ninja.[br]";
         $player->pvp_losses++;
         $player->last_pvp_ms = System::currentTimeMs();
         $player->last_death_ms = System::currentTimeMs();
+        $player->pvp_immunity_ms = System::currentTimeMs() + (5 * 60 * 1000); // 5 minutes
```
levimeahan, line 191: Optionally we could put this value as a constant in a class somewhere, for easier configuration

levimeahan (approved):
Sweeeeeet! Looks good

cron_jobs/region_cron.php
aumu1031: Agreed, we're going to need some more jobs probably, like monthly tracking (pvp, points, etc), so standardizing would be great.

pages/battle.php
aumu1031: Any preference on where it would belong? `User` probably.

PR #437, opened 2023-10-01
War Phase 3, VillageHQ/Kage

aumu1031:


classes/Village.php
```diff
@@ -129,4 +135,287 @@ public function getRelations(System $system, string $village_id): array {
...
+    public static function claimSeat(System $system, User $player, string $seat_type): string {
+        switch ($seat_type) {
+            case 'kage':
+                // check seat available
+                $result = $system->db->query("SELECT * FROM `village_seats` WHERE `village_id` = {$player->village->village_id} AND `seat_type` = '{$seat_type}' AND `seat_end` IS NULL LIMIT 1");
+                if ($system->db->last_num_rows > 0) {
+                    return "Seat is occupied!";
+                }
+                // check tier
+                if (!$player->reputation->rank >= 5) {
```
levimeahan, line 184:
```suggestion
                if ($player->reputation->rank < Village::MIN_KAGE_REP_TIER) {
```

classes/Village.php
```diff
@@ -129,4 +135,287 @@ public function getRelations(System $system, string $village_id): array {
...
+    {
+        $villager_result = $system->db->query("SELECT COUNT(*) AS count FROM `users` WHERE `village` = '{$village}'");
+        $villager_result = $system->db->fetch($villager_result);
+        return $villager_result['count'];
+    }
+
+    /**
+     * @return bool
+     */
+    public static function claimSeat(System $system, User $player, string $seat_type): string {
```
levimeahan, line 175: Could you extract this logic function to a `VillageManager` class so the `Village` class itself can be closer to a static data object?

I think other "write" functions like `updateResources` are good candidates as well

classes/Village.php
```diff
@@ -129,4 +135,287 @@ public function getRelations(System $system, string $village_id): array {
...
+        switch ($seat_type) {
+            case 'kage':
+                // check seat available
+                $result = $system->db->query("SELECT * FROM `village_seats` WHERE `village_id` = {$player->village->village_id} AND `seat_type` = '{$seat_type}' AND `seat_end` IS NULL LIMIT 1");
```
levimeahan, line 179: I wonder if it's worth saving the current kage in a single field in the `Village` class so there's always one authoritative "this is the kage" value.

Hypothetically as long as village operations are wrapped in transactions we shouldn't end up in a state where there are two `village_seats` rows with no end, but having it in one column completely prevents that. We could still use this table to track past kages and how long the current kage has been in seat

classes/Village.php
```diff
@@ -129,4 +135,287 @@ public function getRelations(System $system, string $village_id): array {
...
+                    $seats['kage'] = [
+                        "seat_key" => 'kage',
+                        "seat_id" => $seat['seat_id'],
+                        "user_id" => $seat['user_id'],
+                        "village_id" => $seat['village_id'],
+                        "seat_type" => $seat['seat_type'],
+                        "seat_title" => $seat['seat_title'],
+                        "seat_start" => $seat['seat_start'],
+                        "user_name" => $seat['user_name'],
+                        "avatar_link" => $seat['avatar_link'],
```
levimeahan, line 269: Hmm could you make a DTO for this? Looks like fields are same as the other seats below, would give us the benefits of static types in whatever code is calling this function

classes/Village.php
```diff
@@ -129,4 +135,287 @@ public function getRelations(System $system, string $village_id): array {
...
+            // can determine resources lost based on difference of produced and collected
+            $resource_history[$resource_id]['lost'] = $resource_history[$resource_id]['produced'] - $resource_history[$resource_id]['collected'];
+            // factor in resources which haven't been collected from regions
+            $result = $system->db->query("SELECT SUM(`resource_count`) as 'quantity' FROM `region_locations` INNER JOIN `regions` on `regions`.`region_id` = `region_locations`.`region_id` WHERE `resource_id` = {$resource_id} AND `village` = {$village_id}");
+            $result = $system->db->fetch($result);
+            $resource_history[$resource_id]['lost'] -= (int) $result['quantity'];
+            // get upkeep - WIP
+            $resource_history[$resource_id]['spent'] = 0;
+        }
+        // extra step to calculate lost resources: factor in resources which haven't been collected from caravans
```
levimeahan, line 360:
```suggestion
        // extra step to calculate lost resources: don't include resources which haven't been collected from caravans as lost
```
I was initially a little confused, does this wording match your intent? Reading the code below it looks like this is what's going on

classes/Village.php
```diff
@@ -129,4 +135,287 @@ public function getRelations(System $system, string $village_id): array {
...
+    /**
+     * @return Village
+     */
+    public static function getVillageByID(System $system, int $village_id): Village
+    {
+        $result = $system->db->query(
+            "SELECT * FROM `villages`
+                WHERE `villages`.`village_id`='{$village_id}'"
+        );
+        $result = $system->db->fetch($result);
```
levimeahan, line 383: We should handle the case of village not loading here (probably by throwing an exception?)

classes/Village.php
```diff
@@ -129,4 +135,287 @@ public function getRelations(System $system, string $village_id): array {
...
+        return $result;
+    }
+
+    /**
+     * @return array
+     */
+    public static function getPlayerSeat(System $system, int $player_id): array
+    {
+        $result = $system->db->query("SELECT * FROM `village_seats` WHERE `user_id` = {$player_id} AND `seat_end` IS NULL LIMIT 1");
+        $result = $system->db->fetch($result);
```
levimeahan, line 403: Maybe we could use the seat info DTO here too?

cron_jobs/caravan_cron.php
```diff
@@ -135,6 +145,9 @@ function hourlyCaravan(System $system, $debug = true): void
             VALUES ('{$start_time}', '{$travel_time}', '{$region_id}', '{$village_id}', '{$caravan_type}', '{$resources}', '{$name}')";
     }
 
+    // all regions_locations should be emptied
+    $queries[] = "UPDATE `region_locations` SET `resource_count` = 0 WHERE `region_id` > 5";
```
levimeahan, line 149: What is the significance of region_id > 5?

cron_jobs/caravan_cron.php
```diff
@@ -144,7 +157,7 @@ function hourlyCaravan(System $system, $debug = true): void
     } else {
         echo "Script running...<br>";
         foreach ($queries as $query) {
-            $system->db->query("LOCK TABLES `caravans` WRITE;");
+            $system->db->query("LOCK TABLES `caravans` WRITE, `region_locations` WRITE, `resource_logs` WRITE, `villages` WRITE;");
```
levimeahan, line 160: Is it possible for other code to read a value from the table while this lock is active, then write it back to the table after this lock has expired?

db/migrations/20230929041633_third_war_migration.php
```diff
@@ -0,0 +1,74 @@
...
+            CREATE TABLE `resource_log` (
+              `log_id` int(11) NOT NULL AUTO_INCREMENT,
+              `village_id` int(11) NOT NULL,
+              `resource_id` int(11) NOT NULL,
+              `type` int(11) NOT NULL,
+              `quantity` int(11) NOT NULL,
+              `time` int(11) NOT NULL,
+              PRIMARY KEY (`id`));
+            )
+        ");
```
levimeahan, line 44: Will be so much easier to run manually if needed, thank you!

ui_components/src/village/Village.js
```diff
@@ -0,0 +1,265 @@
...
+                days: resourceDays.current,
+            }
+        ).then((response) => {
+            if (response.errors.length) {
+                handleErrors(response.errors);
+                return;
+            }
+            switch (resourceDays.current) {
+                case 1:
+                    resourceDays.current = 7;
```
levimeahan, line 34: Little confused, why do we change to 7 if the current value is 1?

aumu1031: On-click it toggled between Daily => Weekly => Monthly => Daily

cron_jobs/caravan_cron.php
aumu1031: I think "LOCK" is a hard lock on read/write from these tables for the duration?

cron_jobs/caravan_cron.php
aumu1031: Regions 1/2/3/4/5 are all home regions of Stone/Cloud/Leaf/Sand/Mist respectively
Since home regions do not have a castle, they don't use caravans. The resource move directly from the villages -> main village during each `region_cron` tick.
An improvement here would be to add a flag on the `regions` table, like `is_home_region`

classes/Village.php
aumu1031: So adding a `kage_id` column to `villages` basically?

api/village.php
```diff
@@ -0,0 +1,69 @@
...
+    } else {
+        throw new RuntimeException('No request was made!');
+    }
+
+    $VillageAPIResponse = new VillageAPIResponse();
+
+    switch($request) {
+        case 'LoadVillageData':
+            $VillageAPIResponse->response = [
+                'policyData' => VillageApiPresenter::boostDataResponse($system, $player),
```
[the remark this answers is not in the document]
aumu1031: Chi is still working on boosts/policies so this is unused for now

classes/Village.php
aumu1031: That makes sense, `Village` class bloated very fast. 
Will move most of the functionality into `VillageManager`.

classes/Village.php
aumu1031: Agreed, will refactor into a DTO.

classes/Village.php
aumu1031: Marking as to-do: add exception handling for all the new village functions.

classes/Village.php
aumu1031: Complete agree, will do.

classes/Village.php
aumu1031: Summary of what's going on
- We know how many resources have been produced
- We know how many resources have been collected
- [Produced - Collected] = Lost, in other words the difference is how many resources other villages have stolen

There are two exceptions to this
- Some resources are still sitting in villages waiting to be collected
- Some resources are currently held in caravans (the comment for the section here)

I'm not 100% sure on how to word this best as a comment, what do you recommend?

api/village.php
```diff
@@ -0,0 +1,69 @@
...
+                'populationData' => VillageApiPresenter::populationDataResponse($$system, $player),
```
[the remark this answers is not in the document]
aumu1031: fixed

classes/Village.php
aumu1031: fixed

api/village.php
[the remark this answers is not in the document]
levimeahan: Should we comment out this line or add a placeholder function returning an empty array then? Seems like this should throw an error since we're trying to call a function that doesn't exist

classes/Village.php
levimeahan: Yup!

classes/Village.php
levimeahan: That makes sense - I think for this part the rewording in my suggestion matches what you described

cron_jobs/caravan_cron.php
levimeahan: Oh you're right I forgot `WRITE` here blocks all access, not just writes. Was thinking there was some kind of "READWRITE" lock 😅

ui_components/src/village/Village.js
levimeahan: Hmm, could we handle that with a wrapper around this function, or rename this to `FetchNextIntervalTypeResources` to be clear on what it's doing?

ui_components/src/village/Village.js
```diff
@@ -0,0 +1,316 @@
...
+    diplomacyData,
+    resourceData,
+    clanData,
+}) {
+    const [seatDataState, setSeatDataState] = React.useState(seatData);
+    const [resourceDataState, setResourceDataState] = React.useState(resourceData);
+    const [playerSeatState, setPlayerSeatState] = React.useState(playerSeat);
+    const [modalState, setModalState] = React.useState("closed");
+    const resourceDays = React.useRef(1);
+    const resourceDaysDisplay = React.useRef("daily");
```
levimeahan, line 20: Curious, why refs here instead of state?

levimeahan (approved):

ui_components/src/village/Village.js
aumu1031: Yeah we can just do state, will change that.

PR #442, opened 2023-10-09
War tweaks

aumu1031:
- Updated war notifications
- Added extra details view to war objectives on hover
- Loot action is twice as fast to prevent lagging behind caravan
- Made loot carrying icons much smaller
- Patrols now take a player's resources on loss
- Added check to make sure operation is valid each progress tick (e.g. region changes control)
- Add extra loot tracking and simplify logic of "lost" resource count
- Allow Jonin to attack Chuunin if carrying loot OR in active war action
- Add damage/reinforce value to travel buttons
- Add max loot carrying capacity (soft limit)


classes/travel/TravelManager.php
```diff
@@ -266,13 +266,26 @@ public function fetchNearbyPlayers(): array {
...
                 && $user['user_id'] != $this->user->user_id
                 && $user['village'] !== $this->user->village->name) {
-                $can_attack = true;
+                if ((int)$user['rank'] === $this->user->rank_num) {
+                    $can_attack = true;
+                } else if ($user['operation'] > 0) {
+                    $can_attack = true;
+                } else if ($loot_count > 0) {
+                    $can_attack = true;
+                }
```
levimeahan, line 288: This gets tricky if we're opening PvP for any war activities, could kinda take away the role of Chuunin - Can you comment this out for now and let's discuss? Would love to get player input on this if we haven't already

classes/village/VillageManager.php
```diff
@@ -304,26 +304,12 @@ public static function getResourceHistory(System $system, int $village_id, int $
             $result = $system->db->fetch($result);
             $resource_history[$resource_id]['claimed'] = (int)$result['claimed'];
             // get lost (lol)
-            // can determine resources lost based on difference of produced and collected
-            $resource_history[$resource_id]['lost'] = $resource_history[$resource_id]['produced'] - $resource_history[$resource_id]['collected'];
-            // factor in resources which haven't been collected from regions
-            $result = $system->db->query("SELECT SUM(`resource_count`) as 'quantity' FROM `region_locations` INNER JOIN `regions` on `regions`.`region_id` = `region_locations`.`region_id` WHERE `resource_id` = {$resource_id} AND `village` = {$village_id}");
+            $result = $system->db->query("SELECT COUNT(*) as 'lost' FROM `loot` WHERE `resource_id` = {$resource_id} AND `target_village_id` = {$village_id} AND (`claimed_village_id` != {$village_id} OR `claimed_village_id` IS NULL) AND (`claimed_time` > {$time} OR `claimed_time` IS NULL)");
```
levimeahan, line 307: Minor, but we could be more clear here

```suggestion
            $result = $system->db->query("SELECT COUNT(*) as 'lost_loot_count' FROM `loot` WHERE `resource_id` = {$resource_id} AND `target_village_id` = {$village_id} AND (`claimed_village_id` != {$village_id} OR `claimed_village_id` IS NULL) AND (`claimed_time` > {$time} OR `claimed_time` IS NULL)");
```

classes/war/Operation.php
```diff
@@ -49,8 +49,10 @@ class Operation
     const BASE_OPERATION_INTERVAL = 12; // time per interval
     const BASE_OPERATION_COST = 150; // chakra/stam cost per interval, 750 total
     /* 100 / 20 * 12s = 60s */
-    const LOOT_GAIN = 2;
-    const LOOT_OPERATION_SPEED = 100; // eat loot action is only 1 interval
+    const LOOT_GAIN = 1;
+    const LOOT_OPERATION_SPEED = 100; // each loot action is only 1 interval
```
levimeahan, line 53: Huh, how does 100 correlate to 1 interval?

classes/war/Operation.php
```diff
@@ -49,8 +49,10 @@ class Operation
...
+    const LOOT_OPERATION_INTERVAL = 6; // takes half time as normal
```
levimeahan, line 54: is this seconds?

classes/war/Operation.php
```diff
@@ -49,8 +49,10 @@ class Operation
...
+    const LOOT_OPERATION_COST = 75; // takes half pool cost
```
levimeahan, line 55: Or we could do "resource", "pool", etc
```suggestion
    const LOOT_OPERATION_ENERGY_COST = 75; // takes half pool cost
```

classes/war/WarManager.php
```diff
@@ -147,20 +157,76 @@ public function beginOperation(int $operation_type, int $target_id, Patrol $patr
...
+            // must be at target location
+            $target_location = new TravelCoords($target['x'], $target['y'], $target['map_id']);
+            if ($this->user->location->fetchString() != $target_location->fetchString()) {
+                return false;
+            }
+        }
+        switch ($operation->type) {
+            case Operation::OPERATION_INFILTRATE:
+                // must be neutral or at war
+                if ($this->user->village->relations[$target['village']]->relation_type != VillageRelation::RELATION_NEUTRAL && $this->user->village->relations[$target['village']]->relation_type != VillageRelation::RELATION_WAR) {
```
levimeahan, line 184: Is there anywhere else we do this? This could be a good candidate for a helper function like `village->villageIsAlly($target['village'])` or `WarManager::villagesAreAllied($this->user->village, $target['village'])`

classes/war/WarManager.php
```diff
@@ -308,4 +386,37 @@ public function handlePatrolDefeat(int $patrol_id) {
...
+            // count loot totals
+            if (empty($loot_gained[$loot['resource_id']])) {
+                $loot_gained[$loot['resource_id']] = 1;
+            } else {
+                $loot_gained[$loot['resource_id']] += 1;
+            }
+        }
+        // update village resources
+        foreach ($loot_gained as $resource_id => $count) {
+            $village->addResource($resource_id, $count);
```
levimeahan, line 414: Hmm - I expect eventually we might want to have controls on how much we add from each type, maybe it would be good to add a type flag to this method like `$village->addResource($resource_id, $count, WarManager::NPC_PATROL_WON)` and then we can control for each way you gather resources, how much it actually adds to the village

But okay to skip, just something to think about for now

pages/war.php
```diff
@@ -54,6 +54,8 @@ function processWarBattleEnd($battle, $player): string {
     $stat_gain_chance = 100;
 
     $battle_result = "";
+    $patrol_id = $battle->getPatrolId();
```
levimeahan, line 57: I've been noodling over what we could call these, thinking maybe "Village Patrol" could be the right term, wdyt?
```suggestion
    $patrol_id = $battle->getVillagePatrolId();
```

pages/war.php
```diff
@@ -117,11 +117,12 @@ function processWarBattleEnd($battle, $player): string {
         $battle_result .= "You have been defeated.";
         $player->health = 5;
         $player->ai_losses++;
-        $player->moveToVillage();
+        //$player->moveToVillage();
         $player->battle_id = 0;
         $player->last_pvp_ms = System::currentTimeMs();
+        $battle_result .= $warManager->handlePatrolWin($patrol_id);
```
levimeahan, line 123: This function is slightly ambiguous to whether the player won or the NPCs did, we may want to name it a little more specifically

templates/villageHQ.php
```diff
@@ -23,7 +23,7 @@
                 seatData: <?= json_encode(VillageApiPresenter::seatDataResponse($system, $player)) ?>,
                 pointsData: <?= json_encode(VillageApiPresenter::pointsDataResponse($system, $player)) ?>,
                 diplomacyData: <?= json_encode(VillageApiPresenter::diplomacyDataResponse($system, $player)) ?>,
-                resourceData: <?= json_encode(VillageApiPresenter::resourceDataResponse($system, $player, 7)) ?>,
+                resourceData: <?= json_encode(VillageApiPresenter::resourceDataResponse($system, $player, 1)) ?>,
```
levimeahan, line 26: What does this number stand for - should we put it in a constant somewhere?

ui_components/src/travel/Travel.js
```diff
@@ -463,13 +463,12 @@ function Travel({
                             )}
                             {((mapData && mapData.operations && !mapData.operation_type) && typeof mapData.operations_type === 'undefined') && (
                                 Object.entries(mapData.operations).map(([key, value], index) => (
-                                    <button key={index} onClick={() => BeginOperation(key)} className='button'>
-                                        {`${value}`}
+                                    <button key={index} onClick={() => BeginOperation(key)} className='button' dangerouslySetInnerHTML={{ __html: value }}>
```
levimeahan, line 466: Double-checking: Where on the backend is this value set, and is it user-generated at all?

levimeahan (approved):

classes/war/Operation.php
aumu1031: 100% progress per interval

classes/war/Operation.php
aumu1031: Interval is in seconds

templates/villageHQ.php
aumu1031: That's just the number of days, we can make explicit when we call the function for clarity. `days: 7`

ui_components/src/travel/Travel.js
aumu1031: Not user generated, this comes from WarManager::getValidOperations
```$health_gain = floor($this->user->level / 2);
$valid_operations[Operation::OPERATION_REINFORCE] .= "<br><span class='reinforce_button_text'>{$health_gain} health</span>";```

File: pages/war.php (hunk 8 above)
Comment (aumu1031) [reply pages/war.php:57]: Sounds good to me!

File: classes/war/WarManager.php (hunk 6 above)
Comment (aumu1031) [reply classes/war/WarManager.php:184]: We also do this elsewhere:
```
if ($this->user->village->village_id == $user['village_id']) {
                $user_alignment = 'Ally';
            }
            else {
                $alignment = $this->user->village->relations[$user['village_id']]->relation_type;
                switch ($alignment) {
                    case VillageRelation::RELATION_NEUTRAL:
                        $user_alignment = 'Neutral';
                        break;
                    case VillageRelation::RELATION_ALLIANCE:
                        $user_alignment = 'Ally';
                        break;
                    case VillageRelation::RELATION_WAR:
                        $user_alignment = 'Enemy';
                        break;
                }
            }
```

File: classes/war/Operation.php (hunk 5 above)
Comment (aumu1031) [reply classes/war/Operation.php:55]: Being more explicit here makes sense.

File: classes/village/VillageManager.php (hunk 2 above)
Comment (aumu1031) [reply classes/village/VillageManager.php:307]: Sounds good to me!

File: classes/travel/TravelManager.php (hunk 1 above)
Comment (aumu1031) [reply classes/travel/TravelManager.php:288]: Will comment out until we discuss the role of Chuunin further.

File: classes/notification/NotificationAPIManager.php [hunk 12]
```diff
@@ -376,17 +376,20 @@ public function getUserNotifications(): array
...
-            " AND `operations`.`type` = " . Operation::OPERATION_RAID);
+            " AND `operations`.`type` = " . Operation::OPERATION_RAID
+            . " GROUP BY `region_locations`.`id`");
         $result = $this->system->db->fetch_all($result);
         foreach ($result as $row) {
             $location = new TravelCoords($row['x'], $row['y'], $row['map_id']);
             $notifications[] = new NotificationDto(
                 action_url: $this->system->router->getUrl('travel'),
-                type: "raid",
+                type: $row['user_village'] == $this->player->village->village_id ? "raid_ally" : "raid_enemy",
```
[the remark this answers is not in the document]
aumu1031: Good idea, we can do this for all notification types.

pages/war.php
aumu1031: `handleLossAgainstPatrol()` or similar

PR #445, opened 2023-10-14
KageWar WIP

aumu1031:


levimeahan (approved):
Approving to unblock testing on dev, will finish reading through changes later

PR #446, opened 2023-10-17
Reputation changes

HitoriWoB:
Added double reputation event
Added +1 static FS reputation

### DIFF
--- a/classes/ForbiddenSeal.php
+++ b/classes/ForbiddenSeal.php
@@ -32,8 +32,8 @@ class ForbiddenSeal {
     public float $extended_training_time;
     public float $extended_training_gains;
     public int $free_transfer_bonus;
-
     public int $max_battle_history_view;
+    public int $bonus_reputation;
 
     /** Display Members (used only in Ancient Market) */
     public string $name_color_display;
@@ -80,6 +80,7 @@ class ForbiddenSeal {
             'extra_stat_transfer_points_per_ak' => 0,
             'max_battle_history_view' => 0,
             'free_transfer_bonus' => 0,
+            'bonus_reputation' => 0,
         ],
         1 => [
             'regen_boost' => 10,
@@ -127,6 +128,7 @@ class ForbiddenSeal {
             'extra_stat_transfer_points_per_ak' => 50,
             'max_battle_history_view' => 10,
             'free_transfer_bonus' => 25,
+            'bonus_reputation' => 0,
         ],
         2 => [
             'regen_boost' => 20, //Report in whole percentages (20 will be .2 bonus)
@@ -174,6 +176,7 @@ class ForbiddenSeal {
             'extra_stat_transfer_points_per_ak' => 100,
             'max_battle_history_view' => 20,
             'free_transfer_bonus' => 50,
+            'bonus_reputation' => 1,
         ]
     );
 
@@ -245,6 +248,7 @@ class ForbiddenSeal {
         $this->extra_stat_transfer_points_per_ak = self::$benefits[$this->level]['extra_stat_transfer_points_per_ak'];
         $this->max_battle_history_view = self::$benefits[$this->level]['max_battle_history_view'];
         $this->free_transfer_bonus = self::$benefits[$this->level]['free_transfer_bonus'];
+        $this->bonus_reputation = self::$benefits[$this->level]['bonus_reputation'];
 
         // Display variables
         $this->name_color_display = self::$benefits[$this->level]['name_color_display'];
--- a/classes/ForbiddenSeal.php
+++ b/classes/ForbiddenSeal.php
@@ -33,7 +33,7 @@ class ForbiddenSeal {
     public float $extended_training_gains;
     public int $free_transfer_bonus;
     public int $max_battle_history_view;
-    public int $bonus_reputation;
+    public int $bonus_pve_reputation;
 
     /** Display Members (used only in Ancient Market) */
     public string $name_color_display;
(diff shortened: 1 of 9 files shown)


aumu1031:
Can you make this also apply the multiplier when getting caps?

aumu1031:
You can reuse the same Event display page by copy/pasting this on the `temp_event.php` page:

```<?php elseif ($system->event instanceof DoubleExpEvent): ?>
<table class="table">
    <tr>
        <th>Event Details</th>
    </tr>
    <tr style="text-align: center">
        <td>Double Experience is active until <?= $system->event->end_time->format("l, M jS, Y")?>.</td>
    </tr>
</table>
<?php else: ?>
    <table class="table">
        <tr><th>Event Details</th></tr>
        <tr><td>There is no active event.</td></tr>
    </table>
<?php endif; ?>```

File: classes/ForbiddenSeal.php [hunk 1]
```diff
@@ -32,8 +32,8 @@ class ForbiddenSeal {
     public float $extended_training_time;
     public float $extended_training_gains;
     public int $free_transfer_bonus;
-
     public int $max_battle_history_view;
+    public int $bonus_reputation;
```
levimeahan, line 36: Let's call this out as `bonus_pve_reputation` to be clear

classes/event/DoubleReputation.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+require_once __DIR__ . '/Event.php';
+
+class DoubleReputation extends Event
```
levimeahan, line 5:
```suggestion
class DoubleReputationEvent extends Event
```
Following convention here

classes/UserReputation.php
```diff
@@ -247,6 +255,14 @@ public function __construct(&$player_rep, &$player_weekly_rep, &$player_pvp_rep,
      * Returns amount of reputation awarded for display/data confirmation purposes
      */
     public function addRep(int $amount, bool $bypass_weekly_cap = false, bool $increment_pvp = false): int {
+        // Double repuation
+        if (!empty($this->event) && $this->event instanceof DoubleReputation) {
+            $amount *= DoubleReputation::rep_modifier;
+        }
+
+        //Bonus seal reputation, 0 by default no need to check, bonus rep amounts are set by seal loading
+        $amount += $this->bonus_rep;
```
levimeahan, line 264: This seems like it would apply to PvP too, right? Per last night I think we want to exclude PvP from these gains

levimeahan (commented):
Most of this looks good, but we need the double rep event to double rep caps as well

classes/UserReputation.php
HitoriWoB: That's correct. Just hadn't had a chance to factor it in.

classes/User.php
```diff
@@ -951,7 +951,7 @@ public function loadTrainingManager(): void
             $this->bloodline_id);
     }
 
-    public function setForbiddenSealFromDb(string $forbidden_seal_db, bool $remote_view) {
+    public function setForbiddenSealFromDb(string $forbidden_seal_db, bool $remote_view, bool $set_seal_bonus) {
```
levimeahan, line 954: An internal flag might be better on this - E.g, the first time we call this, set the bonuses and set `$this->seal_bonuses_applied = true`, the second time, we check the `seal_bonuses_applied` flag and skip applying them (or override the old ones)

classes/event/DoubleReputationEvent.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+require_once __DIR__ . '/Event.php';
+
+class DoubleReputationEvent extends Event
+{
+    const rep_modifier = 2;
+    const pve_modifier = 2;
+    const pvp_modifier = 1;
```
levimeahan, line 9:
```suggestion
    const rep_gain_multiplier = 2;
    const pve_cap_multiplier = 2;
    const pvp_cap_multiplier = 2;
```

classes/UserReputation.php
```diff
@@ -247,8 +261,16 @@ public function __construct(&$player_rep, &$player_weekly_rep, &$player_pvp_rep,
      * Returns amount of reputation awarded for display/data confirmation purposes
      */
     public function addRep(int $amount, bool $bypass_weekly_cap = false, bool $increment_pvp = false): int {
+        // Double repuation
+        if (!empty($this->event) && $this->event instanceof DoubleReputationEvent) {
+            $amount *= DoubleReputationEvent::rep_modifier;
```
levimeahan, line 266:
```suggestion
            $amount *= DoubleReputationEvent::rep_gain_multiplier;
```

classes/UserReputation.php
```diff
@@ -232,6 +240,12 @@ public function __construct(&$player_rep, &$player_weekly_rep, &$player_pvp_rep,
         $this->weekly_cap = $REP_RANK['weekly_cap'];
         $this->weekly_pvp_cap = $REP_RANK['weekly_pvp_cap'];
         $this->base_pvp_reward = $REP_RANK['base_pvp_rep_reward'];
+
+        // EVENT MODIFICATIONS
+        if(!empty($this->event) && $this->event instanceof DoubleReputationEvent) {
+            $this->weekly_cap *= DoubleReputationEvent::pve_modifier;
+            $this->weekly_pvp_cap *= DoubleReputationEvent::pvp_modifier;
```
levimeahan, line 247:
```suggestion
            $this->weekly_cap *= DoubleReputationEvent::pve_cap_multiplier;
            $this->weekly_pvp_cap *= DoubleReputationEvent::pvp_cap_multiplier;
```

levimeahan (approved):

classes/event/DoubleReputationEvent.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+require_once __DIR__ . '/Event.php';
+
+class DoubleReputationEvent extends Event
+{
+    const rep_gain_multiplier = 2;
+    const pve_cap_multiplier = 2;
+    const pvp_cap_multiplier = 1;
```
levimeahan, line 9:
```suggestion
    const pvp_cap_multiplier = 2;
```

levimeahan (approved):

PR #448, opened 2023-10-25
Added Eight Trigram

HitoriWoB:
Added Eight Trigram seal

Benefits:
+25% Regeneration rate
Longer journal (3500 characters)
Journal image size of 600x600
Enhanced long trainings (1.5x length, 2.2x gains)
Enhanced extended trainings (2x length, 3x gains)
Faster stat transfers(+6/minute)
View logs of your last 50 battles

Comparison to T2 seal
+5% Regeneration rate
+1000 journal characters
+100x100 journal image size
Long trainings +20% gains
Extended trainings +25% length, +50% gains
+1 stat/minute transfer
+30 battle log history

Update comes with a seal sale, ends Nov 30:
Upgrading to T3 seal from T1 or T2 will provide a 50% refund, if any credit exists over the purchase price. Estimated refunds will be displayed in the purchase confirmation screen.

### DIFF
--- a/classes/ForbiddenSeal.php
+++ b/classes/ForbiddenSeal.php
@@ -46,7 +46,8 @@ class ForbiddenSeal {
     public static array $forbidden_seal_names = array(
         0 => '',
         1 => 'Twin Sparrow Seal',
-        2 => 'Four Dragon Seal'
+        2 => 'Four Dragon Seal',
+        3 => 'Eight Trigram Seal'
     );
     public static array $benefits = array (
         0 => [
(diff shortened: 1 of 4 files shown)


classes/ForbiddenSeal.php
```diff
@@ -177,6 +178,54 @@ class ForbiddenSeal {
             'max_battle_history_view' => 20,
             'free_transfer_bonus' => 50,
             'bonus_pve_reputation' => 1,
+        ],
+        3 => [
+            'regen_boost' => 20, //Report in whole percentages (20 will be .2 bonus)
```
levimeahan, line 183:
```suggestion
            'regen_boost' => 25, //Report in whole percentages (20 will be .2 bonus)
```

classes/PremiumShopManager.php
```diff
@@ -42,6 +45,30 @@ public function __construct(System $system, User $player) {
...
+    public function loadTierThreeSalePrices($current_seal_level): void {
+        if($this->tierThreeSaleActive() && ($current_seal_level == 0 || $current_seal_level == 3)) {
+            $this->costs['forbidden_seal'][3][60] -= 5;
+            $this->costs['forbidden_seal'][3][90] -= 15;
+        }
+        else {
+            $this->costs['forbidden_seal'][3][60] = $this->costs['forbidden_seal_monthly_cost'][3] * 2;
+            $this->costs['forbidden_seal'][3][90] = $this->costs['forbidden_seal_monthly_cost'][3] * 3;
+        }
+    }
```
levimeahan, line 70: Hmm I think we could probably skip the temporary sale and make these the permanent prices? At least the 3 months discount

HitoriWoB: I think that's reasonable. Should we apply similar discounts to the other two tiers?
Should I leave the 50% refund in place as it is? I think we should give some incentive to switch if someone has 6months of Tier 2

levimeahan: Wait, I realized - don't we already have a "apply AK credit from existing seal to a new seal" in place? 
image

levimeahan: Yeah let's leave that refund in 👍

levimeahan: For discounts to other tiers - Let's leave it at T3 for now, add the others later when we change AK:USD from a 2:1 to 4:1 ratio

classes/ForbiddenSeal.php
```diff
@@ -46,7 +46,8 @@ class ForbiddenSeal {
     public static array $forbidden_seal_names = array(
         0 => '',
         1 => 'Twin Sparrow Seal',
-        2 => 'Four Dragon Seal'
+        2 => 'Four Dragon Seal',
+        3 => 'Eight Trigram Seal'
```
levimeahan, line 50:
```suggestion
        3 => 'Eight Deities Seal'
```

classes/ForbiddenSeal.php
levimeahan, line 183:
```suggestion
            'regen_boost' => 25, //Report in whole percentages (20 will be .2 bonus)
```

levimeahan (approved):
LGTMMMMM

PR #455, opened 2023-10-30
War Tweaks

aumu1031:
- Removed 60s immunity from being attacked after a loss
We already have the 5-min immunity so this is doubling up, and it causes issues in war when someone does war actions during the immunity.
- Added battle links for old layouts
- Fixed caravan name on region change

### DIFF
--- a/cron_jobs/region_cron.php
+++ b/cron_jobs/region_cron.php
@@ -109,7 +109,7 @@ function hourlyRegion(System $system, $debug = true): void
             // if one of the home regions, collect resources bypassing caravans
             $production = WarManager::BASE_RESOURCE_PRODUCTION;
             if ($region['region_id'] <= 5) {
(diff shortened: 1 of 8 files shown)


cron_jobs/region_cron.php
```diff
@@ -109,6 +109,7 @@ function hourlyRegion(System $system, $debug = true): void
             // if one of the home regions, collect resources bypassing caravans
             $production = WarManager::BASE_RESOURCE_PRODUCTION;
             if ($region['region_id'] <= 5) {
+                //$production += WarManager::HOME_REGION_RESOURCE_BONUS;
```
levimeahan, line 112: Should this be uncommented?

levimeahan (approved):

cron_jobs/region_cron.php
aumu1031: Undoing this for now.

PR #459, opened 2023-11-02
Blacklist & Lockout QoL

HitoriWoB:
**Blacklist Update**
Blacklist now works in Inbox (direct messages) as it does in chat, by hiding blocked users in pre-existing multi-chats.
Note: This may lead to "ghost" notifications should someone on your BL post in the group DM.

**Account Lockout Change**
Account lockouts will now function differently than they did in the past.
Failing to login from new location 3 or more times will result in a 5 minute cooldown period before you can login again (this will stack to 8 - 10 minutes with two additional fails in a 10 minute period).
Failing to login from a previous location (last two recorded) will result in a 3 minute cooldown period before you can login again.
Head Moderators and User Admins can still manually unlock accounts to bypass this cooldown period.

### DIFF
--- a/classes/User.php
+++ b/classes/User.php
@@ -95,7 +95,7 @@ class User extends Fighter {
     public int $user_id;
     public string $user_name;
     public int $free_username_changes;
-    public $blacklist;
+    public Blacklist $blacklist;
     public $original_blacklist;
 
     /** @var DailyTask[] */
@@ -466,22 +466,10 @@ class User extends Fighter {
         $this->censor_explicit_language = (bool)$user_data['censor_explicit_language'];
 
         // Message blacklist
-        $this->blacklist = [];
-        $result = $this->system->db->query(
-            "SELECT `blocked_ids` FROM `blacklist` WHERE `user_id`='$this->user_id' LIMIT 1"
+        $this->blacklist = new Blacklist(
+            system: $this->system,
+            user_id: $this->user_id
         );
-        if($result->num_rows != 0) {
-            $blacklist = $this->system->db->fetch($result);
-            $this->blacklist = json_decode($blacklist['blocked_ids'], true);
-            $this->original_blacklist = $this->blacklist;
-        }
-        else {
-            $blacklist_json = json_encode($this->blacklist);
-            $this->system->db->query(
-                "INSERT INTO `blacklist` (`user_id`, `blocked_ids`) VALUES ('{$this->user_id}', '{$blacklist_json}')"
-            );
-            $this->original_blacklist = []; // Default an empty array, user did not have an original.
-        }
 
         // Rank stuff
         $this->rank_num = $user_data['rank'];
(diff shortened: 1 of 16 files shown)


classes/User.php
```diff
@@ -466,22 +470,10 @@ public function loadData($UPDATE = User::UPDATE_FULL, $remote_view = false): voi
         $this->censor_explicit_language = (bool)$user_data['censor_explicit_language'];
 
         // Message blacklist
-        $this->blacklist = [];
-        $result = $this->system->db->query(
-            "SELECT `blocked_ids` FROM `blacklist` WHERE `user_id`='$this->user_id' LIMIT 1"
+        $this->blacklist = new Blacklist(
+            system: $this->system,
+            user_id: $this->user_id
         );
```
levimeahan, line 476: Love it

db/migrations/20231102052648_blacklist_misc_changes.php
```diff
@@ -0,0 +1,29 @@
...
+
+use Phinx\Migration\AbstractMigration;
+
+final class BlacklistMiscChanges extends AbstractMigration
+{
+    public function up(): void {
+        $this->execute("
+            ALTER TABLE `chat` 
+                ADD `user_id` INT(11) NOT NULL DEFAULT 0 AFTER `post_id`,
+                ADD `deleted` SMALLINT(2) NOT NULL DEFAULT 0 AFTER `edited`;
```
levimeahan, line 12: We need to add an index on `deleted` as well since it will be in the `WHERE` clause of every chat refresh query

classes/inbox/Inbox.php
```diff
@@ -140,27 +137,31 @@ public static function getUserData(System $system, string $user_name): ?InboxUse
...
         $user_ids = [];
         foreach($convo_members as $member) {
             $user_ids[$member->user_id] = $member->user_id;
         }
         foreach($convo_members as $member) {
-            if (empty($member->blocked_ids)) {
+            $blacklist = new Blacklist(
+                system: $system,
+                user_id: $member->user_id
+            );
```
levimeahan, line 157: To consider - This will trigger an individual database query for every member of the conversation. How many times might `checkBlacklist` be called on one page load or API request?

classes/user/Blacklist.php
```diff
@@ -0,0 +1,115 @@
...
+class Blacklist {
+    public function __construct(
+        public System $system,
+        public int $user_id,
+        public string $encodedBlacklist = "[]",
+        public array $blacklist = array(),
+        public bool $update = false
+    ) {
+        $this->loadBlacklist();
+    }
```
levimeahan, line 11: If the class needs DB info to initialize, I kind of like the pattern to make a static function to do that like `Blacklist::fromUserId`, and make the constructor not trigger any DB queries. This has the advantage that if you want to optimize later and do a batch query for a bunch of blacklists, you can just pass each row of blacklist data into the constructor

levimeahan (commented):
This blacklist refactor is so satisfying 🙌 

Only blocking comment is the index on `deleted`, then this is good to go

classes/inbox/Inbox.php
HitoriWoB: Disabling for now, I want to rethink my approach.

classes/user/Blacklist.php
HitoriWoB: Will include this with the Inbox update

HitoriWoB:
This is ready for final review. Blocking comment resolved, will have final solution with character PR.

levimeahan (approved):

PR #460, opened 2023-11-02
War Tweaks 2

aumu1031:
- Fix town looting bug
- Add protection for Chuunin by attack-able Jonin allies at same location
- Add ability for either village to force peace after 7 days
- Add 3-day war cooldown for each pair of villages
- Add +1 Stealth bonus during Special Missions (restrict Arena-access and must start in village)
- Rework PvP immunity/delays
~~Now uses the `pvp_immunity_ms` value for PvP cooldowns~~
~~Lasts 90s down from 5 minutes~~
Cannot attack other players during the 90s (up from 60s)
Cannot start a war action for 90s
~~Both players receive pvp immunity/restriction on a draw~~
- PvP wins on a town or castle increases or decreases the defense value based on player/opponent relation to the owner
- Ally towns are no longer visible outside of scout range
- ~~Town/Castle defense now shifts by 5/hour toward baseline (up from 1/hour)~~
- Defeating patrols decreases region defense by 1
- Misc: travel errors now display even if identical to the previous message
- Misc: potential fix for server lag (?) refactor battle refresh

image
image


classes/travel/TravelManager.php
```diff
@@ -287,6 +291,12 @@ public function fetchNearbyPlayers(): array {
                     $can_attack = true;
                 }
             }
+            // additionally, if player is jonin and target is chuunin
+            $is_protected = false;
+            if (User::isProtectedByAlly($this->system, user_id: $user['user_id']) && $this->user->rank_num == 4) {
```
levimeahan, line 296: This is going to trigger <number of nearby players> database queries on every scout API request and we do those for every player viewing the travel page every 400ms, so this could become, worst case if you have 30 players online who all can see each other viewing scout page - 2,250 extra queries per second hitting the database.

I think we need to make this a batch query

classes/war/WarManager.php
```diff
@@ -434,6 +439,21 @@ public function handleWinAgainstPatrol(int $patrol_id) {
...
+            INNER JOIN `regions` on `regions`.`region_id` = `region_locations`.`region_id`
+            WHERE `region_locations`.`region_id` = {$patrol_result['region_id']}
+        ");
+        $location_result = $this->system->db->fetch_all($location_result);
+        foreach ($location_result as $location) {
+            if ($location['defense'] > 0 && $location['village'] == $patrol_result['village_id']) {
+                $location['defense']--;
+                $this->system->db->query("UPDATE `region_locations` SET `defense` = {$location['defense']} WHERE `id` = {$location['id']}");
+            }
+        }
```
levimeahan, line 454: Could be a helper function like `decreaseRegionLocationDefenses()`

ui_components/src/travel/Travel.js
```diff
@@ -255,6 +255,7 @@ function Travel({
 
     function handleErrors(errors) {
         console.warn(errors);
+        setFeedback(null);
```
levimeahan, line 258: Add a comment about why this is needed

levimeahan (approved):

PR #466, opened 2023-11-09
War Table WIP

aumu1031:
image

Queries to fix region capture count on prod:
```
UPDATE player_war_logs AS total
SET total.regions_captured = (
    SELECT SUM(war.regions_captured)
    FROM player_war_logs AS war
    WHERE war.user_id = total.user_id AND war.relation_id IS NOT NULL
)
WHERE total.relation_id IS NULL;
UPDATE village_war_logs AS total
SET total.regions_captured = (
    SELECT SUM(war.regions_captured)
    FROM village_war_logs AS war
    WHERE war.village_id = total.village_id AND war.relation_id IS NOT NULL
)
WHERE total.relation_id IS NULL;
```

PvP Win: 100 Battle, 10 Objective (from Defense)

Patrol Win: 25 Battle, 30 Objective (from Defense)

Region Capture: 1500 Objective

Village Capture: 500 Objective

Reinforce: 
 - 10-35 Objective
 - 250-875/hour

Infiltrate: 
- 10 Objective, 0-37.5 Resource
- 250-1187/hour

Raid: 
- 10-60 Objective
- 250-1500/hour

Loot: 
- 75 Resource
- 1500/caravan


api/village.php
```diff
@@ -163,6 +163,12 @@
                 'challengeData' => VillageManager::getChallengeData($system, $player),
             ];
             break;
+        case 'GetGlobalWarLeaderboard':
+            $page_number = $system->db->clean($_POST['page_number']);
```
levimeahan, line 167:
```suggestion
            $page_number = (int)$_POST['page_number'];
```

classes/village/VillageApiPresenter.php
```diff
@@ -116,4 +116,45 @@ function (ChallengeRequestDto $challenge) {
             VillageManager::getChallengeData($system, $player)
         );
     }
+    public static function warLogDataResponse(System $system, User $player, int $page_number = 1): array {
+        $global_leaderboard_war_logs = array_map(
+            function (WarLogDto $war_log) {
```
levimeahan, line 121: I love DTOs
all_of_the

classes/war/WarLogDto.php
```diff
@@ -0,0 +1,38 @@
...
+    public int $points_gained = 0;
+    public int $war_score = 0;
+    public int $objective_score = 0;
+    public int $resource_score = 0;
+    public int $battle_score = 0;
+
+    public function __construct(array $row, string $log_type) {
+        foreach ($row as $key => $value) {
+            $this->$key = $value;
+        }
```
levimeahan, line 35: I know it's more boilerplate but I think we should avoid this pattern wherever possible, and write out the lines to set each value from the array individually.

If we have this magic that auto-sets the values it can hide issues where a new DB column is added but isn't in code, or there's a typo on a column, etc

Could structure this similar to a pattern used elsewhere:

```php
class WarLogDto {
  public function __construct(
    public int $log_id,
    public string $log_type,
    // ... 
  ) {}

  public static function fromDbArray(array $db_row, string $log_type): WarLogDto {
    return new WarLogDto(
      log_id: $db_row['log_id'],
      log_type: $log_type,
    );
  }
}
```

classes/war/WarLogManager.php
```diff
@@ -44,8 +61,10 @@ public static function logRegionCapture(System $system, User $player, int $regio
         $system->db->query("INSERT INTO `region_logs` (`region_id`, `previous_village_id`, `new_village_id`, `user_id`, `capture_time`, `relation_id`) VALUES ({$region_id}, {$region['village']}, {$player->village->village_id}, {$player->user_id}, {$time}, {$player->village->relations[$region['village']]->relation_id})");
         // log for player
         self::updatePlayerLog($system, $player, 1, self::WAR_LOG_REGIONS_CAPTURED, $player->village->relations[$region['village']]->relation_id);
+        self::updatePlayerLog($system, $player, 1, self::WAR_LOG_REGIONS_CAPTURED, null);
```
levimeahan, line 64: What is the `1` here?

layout/_common.php
```diff
@@ -36,6 +36,8 @@ function coreHeading(string $stylesheet_path): string {
 	<script type='text/javascript' src="./scripts/timer.js"></script>
     <script type='text/javascript' src='./scripts/react-transition-group.js'></script>
     <script type='text/javascript' src='./scripts/luxon.min.js'></script>
+    <script type='text/javascript' src="./scripts/prop-types.min.js"></script>
```
levimeahan, line 39: Where are we using prop-types?

levimeahan (approved):

classes/war/WarLogManager.php
aumu1031: 1 region capture

PR #468, opened 2023-11-10
Misc update

HitoriWoB:
**NavigationAPIManager changes**
Updated to use static method for construction
Added routes as a member
Updated to use consolidated method to return side menu links
Added support for dev-only side menu links

**User Reputation Changes**
Updated pvp win method to fix issue that resulted in ghost reputation appearing to have been awarded
(special thinks to Arth for finding that!)

**Notification Settings**
It is now possible to filter some notifications in settings page.

### DIFF
--- a/db/migrations/20231109235138_misc_updates.php
+++ b/db/migrations/20231109235138_misc_updates.php
@@ -1,0 +1,20 @@
+<?php
+declare(strict_types=1);
+
+use Phinx\Migration\AbstractMigration;
+
+final class MiscUpdates extends AbstractMigration
+{
+    public function up(): void
+    {
+        $this->execute("
+            ALTER TABLE `users` ADD COLUMN `blocked_notifications` TEXT NOT NULL;
+        ");
+    }
+
+    public function down(): void {
+        $this->execute("
+            ALTER TABLE `users` DROP `blocked_notifications`;
+        ");
+    }
+}
--- a/api/navigation.php
+++ b/api/navigation.php
@@ -22,7 +22,7 @@ try {
     }
 
     $NavigationResponse = new NavigationAPIResponse();
-    $NavigationManager = new NavigationAPIManager($system, $player);
+    $NavigationManager = NavigationAPIManager::loadNavigationAPIManager($system, $player);
 
     switch ($request) {
         case "getNavigationLinks":
--- a/classes.php
+++ b/classes.php
@@ -50,6 +50,7 @@ require_once __DIR__ . "/classes/notification/NotificationAPIManager.php";
 require_once __DIR__ . "/classes/notification/NotificationAPIPresenter.php";
 require_once __DIR__ . "/classes/notification/NotificationAPIResponse.php";
 require_once __DIR__ . "/classes/notification/NotificationManager.php";
+require_once __DIR__ . "/classes/notification/BlockedNotificationManager.php";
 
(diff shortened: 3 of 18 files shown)


HitoriWoB:
This is now ready for review @levimeahan

classes/news/NewsManager.php
```diff
@@ -40,10 +40,20 @@ public function saveNewsPost(NewsPostDto $newsPost): bool
...
                 foreach ($user_ids as $user) {
+                    // Blocked notification
+                    $blockedNotifManager = BlockedNotificationManager::BlockedNotificationManagerFromDb(
+                        system: $this->system,
+                        blocked_notifications_string: $user['blocked_notifications']
+                    );
+                    if($blockedNotifManager->notificationBlocked(NotificationManager::NOTIFICATION_NEWS)) {
+                        continue;
+                    }
+                    echo "sending notif to {$user['user_id']}<br />";
```
levimeahan, line 55: Change this echo to debug or a log

db/migrations/20231109235138_misc_updates.php
```diff
@@ -0,0 +1,20 @@
...
+declare(strict_types=1);
+
+use Phinx\Migration\AbstractMigration;
+
+final class MiscUpdates extends AbstractMigration
+{
+    public function up(): void
+    {
+        $this->execute("
+            ALTER TABLE `users` ADD COLUMN `blocked_notifications` TEXT NOT NULL;
```
levimeahan, line 11: starting to think we should really split a lot of the fields from users table into separate tables, but not necessarily something to do in this PR

pages/premium_shop.php
```diff
@@ -24,8 +24,9 @@ function premiumShop(): void {
                 renderPurchaseConfirmation(
                     purchase_type: "user_reset",
                     confirmation_type: "confirm_reset",
-                    confirmation_string: "Are you sure you want to reset your character? You will lose all your stats,
-                bloodline, rank and clan. You will keep your money.",
+                    confirmation_string: "Are you sure you want to reset your character?<br />
+                        You will lose all your stats, bloodline, rank and clan. You will keep your money.<br />
+                        Any active stat transfers will be cancelled and any AK spent will be lost.",
```
levimeahan, line 29: Could we make this conditional if the player has an active stat transfer?

ui_components/src/profile/Profile.js
```diff
@@ -44,6 +46,75 @@ function Profile({
...
+            let secondsString = ((seconds < 10) ? "0" + seconds : seconds);
+
+            returnString += secondsString;
+        }
+        else {
+            returnString += "00";
+        }
+
+        return returnString;
+    }
```
levimeahan, line 106: Could this function be moved out of this component? It doesn't look like it relies on any of the variables inside the component.

It could still be in this file, but maybe down below

levimeahan (approved):
Couple comments to address but overall looks good! Awesome upgrades, blocked notification system will be great

db/migrations/20231109235138_misc_updates.php
HitoriWoB: I agree. I'll throw that in with characters.

classes/news/NewsManager.php
HitoriWoB: Whoopos, meant to delete that

PR #473, opened 2023-11-13
Fix Ashes Policy

aumu1031:


classes/village/VillageManager.php
```diff
@@ -1284,15 +1284,15 @@ public static function enactProposal(System $system, User $player, int $proposal
                 // policy restriction - check not in alliance
                 if (VillagePolicy::$POLICY_EFFECTS[$proposal['policy_id']][VillagePolicy::POLICY_RESTRICTION_ALLIANCE_ENABLED] == false) {
                     $relation_type = VillageRelation::RELATION_ALLIANCE;
-                    $system->db->query("SELECT COUNT(*) FROM `village_relations` WHERE `relation_end` IS NULL AND `relation_type` = {$relation_type} AND (`village1_id` = {$proposal['village_id']} OR `village2_id` = {$proposal['village_id']})");
+                    $system->db->query("SELECT * FROM `village_relations` WHERE `relation_end` IS NULL AND `relation_type` = {$relation_type} AND (`village1_id` = {$proposal['village_id']} OR `village2_id` = {$proposal['village_id']})");
                     if ($system->db->last_num_rows > 0) {
```
levimeahan, line 1288: QQ, would this work? I'm okay with this as is, not a big perf hit here, question would only be about which pattern we want to use
```suggestion
                    $result = $system->db->query("SELECT COUNT(*) as `count` FROM `village_relations` WHERE `relation_end` IS NULL AND `relation_type` = {$relation_type} AND (`village1_id` = {$proposal['village_id']} OR `village2_id` = {$proposal['village_id']})");
                    $count = $system->db->fetch($result)['count']
                    if ($count > 0) {
```

levimeahan (approved):

classes/village/VillageManager.php
aumu1031: We use `SELECT * FROM TABLE LIMIT 1` quite a bit already, here I tried to be more direct and check `COUNT` but mixed up the two patterns and caused the original bug in the process.
I kind of wish we had a helper for this to force consistency, like `$system->db->last_has_results()` which works for both query patterns (SELECT *, and SELECT COUNT(*)). Not sure if that would work though.

levimeahan: Yeah, a helper would be great. Hmm... Will give this more thought

PR #474, opened 2023-11-14
Region + Resource Trading

aumu1031:
image


classes/village/VillageProposalDto.php
```diff
@@ -14,6 +14,7 @@ public function __construct(
         public ?int $policy_id,
         public ?string $vote_time_remaining,
         public ?string $enact_time_remaining,
+        public array $trade_data = [],
```
levimeahan, line 17: I think this should really be its own DTO, it's not a variable-length list of things, it's used as a structured data object. Especially as there are multiple levels of nested, labeled fields here and it's hard to track down all of them

classes/village/VillageManager.php
```diff
@@ -1498,6 +1528,49 @@ public static function enactProposal(System $system, User $player, int $proposal
                     SET `region_locations`.`occupying_village_id` = NULL
                     WHERE `regions`.`village` = {$proposal['village_id']} AND `region_locations`.`occupying_village_id` = {$proposal['target_village_id']}");
                 break;
+            case self::PROPOSAL_TYPE_OFFER_TRADE:
+                // check at war
+                if ($player->village->isEnemy($proposal['target_village_id'])) {
```
levimeahan, line 1533: Hmm, should we explicitly make this allies only?

classes/village/VillageManager.php
```diff
@@ -1826,12 +1899,142 @@ public static function createBreakAllianceProposal(System $system, User $player,
...
+        $village2_regions = $system->db->query("SELECT `region_id` FROM `regions` WHERE `village` = {$village2_id}");
+        $village2_regions = $system->db->fetch_all($village2_regions);
+        // get resources
+        $village1_resources = self::getResources($system, $village1_id);
+        $village2_resources = self::getResources($system, $village2_id);
+        // check valid
+        foreach ($offered_resources as $resource) {
+            $resource_id = $resource['resource_id'];
+            $required_count = $resource['count'];
+            if (!isset($village1_resources[$resource_id]) || $village1_resources[$resource_id] < $required_count) {
```
levimeahan, line 1957: Kind of worried that someone can trade away all their resources, this seems like one of the biggest potentials for abuse. We do have other safeguards around getting Kage powers, but maybe it's best to put a cap on # of resources you can do in one trade? e.g. 10-20% of village's total resources?

levimeahan (commented):
Awesome update! I do have a couple concerns about abusability. let me know what you think

classes/village/VillageProposalDto.php
aumu1031: Yeah that makes sense, while developing it was easier to just grab what I needed while experimenting but especially as this grows to more complex trades, nested arrays will get even more messy. The next time we touch trading (e.g. adding the ability to create counter-offers) I'll rework this to a DTO.

classes/village/VillageManager.php
aumu1031: Yeah I'm not sure why I wanted the ability to trade with neutral.

levimeahan (approved):

classes/village/VillageManager.php
aumu1031: Wondering if it should be a flat value versus %, but otherwise agree on the restriction.

Issue #129, opened 2021-10-10
Tasks completed tracker


unownram:
Like something that counts how many daily tasks have been completed by the user. Can have multiple things like easy, medium, hard, total missions completed

HitoriWoB:
Mission tracking has been added. I don't know that full daily mission tracking is a good approach. We could perhaps make completing 100 daily tasks an achievement (or the like).

PR #475, opened 2023-11-14
Special Mission tracking

HitoriWoB:
Special missions now award mission completes

Genin => D-rank
Chuunin => C-rank
Jonin => B-Rank
Sennin => A-Rank

Selecting an easy difficulty at any rank will downgrade the mission 1-rank
Selecting a hard or nightmare difficulty at any rank will upgrade the mission 1-rank

Being a capped Chuunin will upgrade the mission 1-rank
Selecting an easy or normal mission as a capped Chuunin will downgrade the mission 1-rank (stacks with base dg on easy)

Being level 90+ Jonin will upgrade the mission 1-rank
Being a level 80+ Jonin and selecting easy or normal difficulty will downgrade the mission 1-rank (stacks with base dg on easy)

These will progress daily missions, so long as the rank meets the minimum daily task requirement

Notes:
SM will always county as a minimum of D-rank
Mission rank awards are hard-capped to A-rank


classes/SpecialMission.php
```diff
@@ -6,6 +6,10 @@ class SpecialMission {
     const DIFFICULTY_HARD = 'hard';
     const DIFFICULTY_NIGHTMARE = 'nightmare';
 
+    // Used for incrementing mission ranks & completing daily tasks
+    public static array $UPGRADE_DIFFICULTIES = [self::DIFFICULTY_HARD, self::DIFFICULTY_NIGHTMARE];
+    public static int $MAX_MISSION_LEVEL = Mission::RANK_A;
```
levimeahan, line 11:
```suggestion
    public static int $MAX_MISSION_RANK = Mission::RANK_A;
```

classes/SpecialMission.php
```diff
@@ -493,6 +497,20 @@ public function completeMission($progress): string {
...
+        if(isset($this->player->missions_completed[$mission_rank])) {
+            $this->player->missions_completed[$mission_rank]++;
+        }
+        else {
+            $this->player->missions_completed[$mission_rank] = 1;
+        }
+        // Check daily tasks
+        if($this->player->daily_tasks->hasTaskType(type: DailyTask::ACTIVITY_MISSIONS)) {
+            $this->player->daily_tasks->progressTask(activity: DailyTask::ACTIVITY_MISSIONS, amount: 1, sub_task: $mission_rank);
+        }
```
levimeahan, line 511: This is cleaaaaan I love it

classes/SpecialMission.php
```diff
@@ -784,6 +802,64 @@ private function calcHealthLost(int|float $base_hp_lost_percent): int {
...
+            }
+        }
+        // Jonin level upgrade, begin at level 90
+        if($this->player->rank_num == 4 && $this->player->level >= 90) {
+            $mission_rank = $this->upgradeMissionRank($mission_rank);
+        }
+        // Jonin level downgrade based on NORMAL difficulty (intended to STACK with easy downgrade above)
+        if($this->player->rank_num == 4 && $this->player->level >= 75 && in_array($this->difficulty, [self::DIFFICULTY_EASY, self::DIFFICULTY_NORMAL])) {
+            $mission_rank = $this->downgradeMissionRank($mission_rank);
+        }
```
levimeahan, line 855: This logic gets a little weird and pushes me towards - should we do the level-based system where we assign a level to each SM and then a level range for each mission rank? I think that's easier to understand

Config array example

```php
const SPECIAL_MISSION_LEVELS = [
   // AS
   1 => [
       self::DIFFICULTY_EASY = 1,
       self::DIFFICULTY_MEDIUM => 3,
       self::DIFFICULTY_HARD => 6,
       self::DIFFICULTY_NIGHTMARE = 9
   ] 
];
```

this also allows us to dynamically tune difficulty to say "this level is when you should be able to clear this mission"

HitoriWoB: Love how clean this is... my brain over-complicates things sometimes ~.~

levimeahan: Don't look at early WoB code, it was pretty gnarly 😅

HitoriWoB:
I believe this should work

levimeahan (approved):

HitoriWoB:
This will resolve #129

PR #478, opened 2023-11-16
Increase trade_data Column Size

aumu1031:


levimeahan (changes requested):
Please do this in a new migration, not by editing the existing one - This way we can ensure consistency across the environments

aumu1031:
> Please do this in a new migration, not by editing the existing one - This way we can ensure consistency across the environments

I fixed this manually on live to prevent players from encountering bugs, so this migration is just for developers.

levimeahan (approved):
Gotcha - I do think it's better to avoid this anyway, but will approve to unblock this instance

PR #482, opened 2023-11-22
Multi-Tag Jutsu

aumu1031:
image
image
image
image
image


classes/Jutsu.php
```diff
@@ -140,7 +146,8 @@ class Jutsu {
      * @param string      $hand_seals
      */
     public function __construct(int $id, string $name, int $rank, string $jutsu_type, float $base_power, int $range,
-        ?string $effect, ?float $base_effect_amount, ?int $effect_length, string $description, string $battle_text, int $cooldown,
+        ?string $effect_1, ?float $base_effect_amount_1, ?int $effect_length_1, ?string $effect_2, ?float $base_effect_amount_2, ?int $effect_length_2,
```
levimeahan, line 149:
```suggestion
        ?Effect $effect_1, ?Effect $effect_2,
```
As a follow-up, we could move to this pattern for the constructor (mapping from DB fields to these can be done in `fromArray()`

levimeahan (approved):

PR #484, opened 2023-11-23
Elemental Weakness

aumu1031:
- Cleanup elemental clash logic
- Add elemental weakness tag: hard set to 20% effect
- Elemental weakness and clash applied in same step, so Fire vs Water against a target with Fire Weakness = 0% damage change.
- To-do: after multi-effect jutsu merge will update effect display


classes/battle/BattleEffectsManager.php
```diff
@@ -498,6 +522,21 @@ public function getAnnouncementText(string $effect) : string{
             case 'cast_speed_boost':
                 $announcement_text = "[player]'s Cast Speed is being increased";
                 break;
+            case 'weakness_fire':
+                $announcement_text = "[opponent] gains weakness to Fire";
```
levimeahan, line 526:
```suggestion
                $announcement_text = "[opponent] is vulnerable to Fire";
```

classes/battle/BattleManager.php
```diff
@@ -853,54 +853,88 @@ public function jutsuCollision(
...
+        // Fire > Wind > Lightning > Earth > Water > Fire
+        $elemental_clash_damage_modifier = 0.2;
+        $player1_elemental_damage_modifier = 1;
+        $player2_elemental_damage_modifier = 1;
+
+        // Calculate player1 elemental damage modifier
+        if (!empty($player1_jutsu->element)) {
+            switch (strtolower($player1_jutsu->element)) {
+                case 'fire':
+                    $player1_elemental_damage_modifier += $player2->fire_weakness;
```
levimeahan, line 865: I wonder if it's more intuitive to do this multiplicatively instead of additively. Conceptually the two jutsu clashing has nothing to do with how resistant one player is to the element, we could apply the clash first and then the weakness second.

e.g. has 50% weakness to fire, but uses a water jutsu which counters the incoming fire jutsu:
- Jutsu start damage 10k
- Jutsu damage *= 80% (water counter) = 8,000
- Jutsu damage *= 150% (fire weakness) = 12,000

as opposed to
- jutsu start damage 10k
- Modifier = 100% - 20% + 50% = 130%
- Jutsu damage *= 130% = 13,000

levimeahan (approved):

classes/battle/BattleManager.php
aumu1031: Multiplication gets weird to me in general because the effect of each multiplier depends on all other multipliers.

Double check my logic on this?

If you're adding bonuses in series, you want to keep them as additive, or else each matters more than the last.
e.g. +50% and +50% could be +100% bonus (additive) or +125% (multiplicative)

If you're subtracting penalties in series, they should be multiplicative, or else each matters more than the last.
e.g. -50% and -50% could be a -100% penalty (additive) or -75% (multiplicative)

If you're adding a bonus and penalty in series, you want to keep them as additive, or else the penalty matters more.
e.g. +20% and -20% could be a 0% change (additive) or -4% (multiplicative)

So I think this depends on player perception. Will players see Element Clash and Elemental Vulnerability as part of the same thing? If so, 50% vulnerability becoming 20% in a clash rather than 30% feels very unintuitive.

classes/battle/BattleEffectsManager.php
aumu1031: Updated

classes/battle/BattleManager.php
levimeahan: The number rationale about penalties/bonuses makes sense - My main opinion here is on #2, where I think if players are trying to figure out how combat works, it's more intuitive to say that jutsu clashing happens first where elemental clash occurs, then the jutsu hits the player and resist/vulnerability is applied. 

Within each of those distinct phases we can do additive penalty/bonuses, but since the two phases are calculated separately, they are multiplicative of each other

Phase 1: Apply jutsu clash effects (bonuses + penalties additive)
Phase 2: Apply elemental resists/weaknesses (e.g. this is in `calcDamageTaken` not in the jutsu clash step)

But I'm okay with either option here if you have a strong opinion

aumu1031: > The number rationale about penalties/bonuses makes sense - My main opinion here is on #2, where I think if players are trying to figure out how combat works, it's more intuitive to say that jutsu clashing happens first where elemental clash occurs, then the jutsu hits the player and resist/vulnerability is applied.
> 
> Within each of those distinct phases we can do additive penalty/bonuses, but since the two phases are calculated separately, they are multiplicative of each other
> 
> Phase 1: Apply jutsu clash effects (bonuses + penalties additive) Phase 2: Apply elemental resists/weaknesses (e.g. this is in `calcDamageTaken` not in the jutsu clash step)
> 
> But I'm okay with either option here if you have a strong opinion

I'm leaning toward putting it in `calcDamageTaken` for now, thanks!

PR #488, opened 2023-11-26
Vulnerability Tag

aumu1031:
Applied during jutsu collision, so specifically affects the jutsu's damage
- Applies to jutsu before effects are created (empowers residuals), same player expectation as Offense Boost
- Does not apply retroactively to residuals already inflicted

We could apply Vulnerability in `calcDamageTaken` instead, but IMO that makes strategy less intuitive.
It would mean its better to apply your residuals first then apply Vulnerability rather than using it as a combo start
It would also mean that its better to use jutsu with higher base power during Vulnerability windows rather than being an overall modifier to that jutsu's potential damage output


levimeahan:
> We could apply Vulnerability in calcDamageTaken instead, but IMO that makes strategy less intuitive.
It would mean its better to apply your residuals first then apply Vulnerability rather than using it as a combo start
It would also mean that its better to use jutsu with higher base power during Vulnerability windows rather than being an overall modifier to that jutsu's potential damage output
To me, this is less intuitive. When the effect is "target is vulnerable to fire" if they are already on fire, I expect them to take more damage from it. And this also creates meaningful strategic difference between a "boost" and a "vulnerability" and allows for interesting and more flexible combos such as boost => residual => residual => vulnerability => immolate.

So I'm fairly strongly in the camp of "apply in `calcDamageTaken`, increasing residual damage taken while active" for vulnerability effects.

aumu1031:
> > We could apply Vulnerability in calcDamageTaken instead, but IMO that makes strategy less intuitive.
> > It would mean its better to apply your residuals first then apply Vulnerability rather than using it as a combo start
> > It would also mean that its better to use jutsu with higher base power during Vulnerability windows rather than being an overall modifier to that jutsu's potential damage output
> 
> To me, this is less intuitive. When the effect is "target is vulnerable to fire" if they are already on fire, I expect them to take more damage from it. And this also creates meaningful strategic difference between a "boost" and a "vulnerability" and allows for interesting and more flexible combos such as boost => residual => residual => vulnerability => immolate.
> 
> So I'm fairly strongly in the camp of "apply in `calcDamageTaken`, increasing residual damage taken while active" for vulnerability effects.

Agreed, making adjustment in combat branch.

PR #505, opened 2023-12-04
Tweaks

aumu1031:
- BL Jutsu scale to +30% at lv100 down from +50%
- BL Jutsu scale to the player's rank (Genin->Chuunin +40%, Chuunin-> Jonin +25%)
- Substitution now ignores Piercing
- Counter damage is now displayed in battle
- Counter now reflects the % of the target's attack back
- Linked jutsu are usable any turn while the initial jutsu is on cooldown
- Elemental clash reduces effect by same %
- Fixed elemental weakness (applies full value not limited to 20%)


levimeahan (approved):
LGTM, some naming stuff but nothing blocking

PR #507, opened 2023-12-05
Combat Tweaks

aumu1031:
- Dev flags
- Code fixes (some manually on dev already)
- Tuned speed ratio value
- Reworked BL jutsu scaling
    - 0.9x Off Skill and 0.6x BL Skill (from 0.8x and 0.8x) 
    - BL builds now tradeoff approximately 10% shop damage in exchange for approximately 20% BL jutsu damage
    - For balanced builds BL jutsu will be similar damage for the same power (e.g. 4.4 shop jutsu = 4.4 BL jutsu)

### DIFF
--- a/classes/battle/BattleManager.php
+++ b/classes/battle/BattleManager.php
@@ -13,7 +13,7 @@ class BattleManager {
     const CAST_SPEED_DAMAGE_REDUCTION_RATIO = 1; // e.g. 10% of your stats in speed = 10% evasion
     const MAX_EVASION_DAMAGE_REDUCTION = 0.5; // LEGACY
     const EVASION_SOFT_CAP = 0.35; // caps at 50% evasion
-    const EVASION_SOFT_CAP_RATIO = 0.65; // evasion beyond soft cap only 50% as effective
+    const EVASION_SOFT_CAP_RATIO = 0.65; // evasion beyond soft cap only 65% as effective
     const EVASION_HARD_CAP = 0.65; // caps at 75% evasion
 
     const RESIST_SOFT_CAP = 0.35; // caps at 35% resist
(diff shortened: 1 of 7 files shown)


classes/battle/BattleManager.php
```diff
@@ -13,15 +13,15 @@ class BattleManager {
     const CAST_SPEED_DAMAGE_REDUCTION_RATIO = 1; // e.g. 10% of your stats in speed = 10% evasion
     const MAX_EVASION_DAMAGE_REDUCTION = 0.5; // LEGACY
     const EVASION_SOFT_CAP = 0.35; // caps at 50% evasion
-    const EVASION_SOFT_CAP_RATIO = 0.5; // evasion beyond soft cap only 50% as effective
+    const EVASION_SOFT_CAP_RATIO = 0.65; // evasion beyond soft cap only 50% as effective
```
levimeahan, line 16:
```suggestion
    const EVASION_SOFT_CAP_RATIO = 0.65; // evasion beyond soft cap only 65% as effective
```
Would this be accurate?

pages/healingShop.php
```diff
@@ -48,13 +48,15 @@ function healingShop() {
 			if(!isset($ramen_choices[$heal])) {
 				throw new RuntimeException("Invalid choice!");
 			}
-			if($player->getMoney() < $ramen_choices[$heal]['cost']) {
+			if($player->getMoney() < $ramen_choices[$heal]['cost'] && !$system->isDevEnvironment()) {
```
levimeahan, line 51: Maybe we could make this simpler by changing the costs above? e.g.


```php
if($system->isDevEnvironment()) {
   $ramen_choices['vegetable']['cost'] = 0;
   $ramen_choices['pork']['cost'] = 0;
   $ramen_choices['deluxe']['cost'] = 0;
}
```

I think this way it would also reflect in the displayed costs

levimeahan (approved):

classes/battle/BattleManager.php
aumu1031: Yep, good catch. I'll change this for the other comments as well.

pages/healingShop.php
aumu1031: Cleaner, changing that now.

PR #508, opened 2023-12-07
Immolate, Recoil, Element Clash, Tag Updates

aumu1031:
- Cleaned up Counter logic (calculated in clash, reduces jutsu power, runs through calcDamageTaken)
- Recoil tag (note: applies after evasion and before barrier, then runs through calcDamageTaken later on)
- Immolate (has two effects: immediately triggers all remaining residuals, and increases jutsu power by a percent of the total)
- Added effect details to combat log e.g. "Arthesia is taking Residual damage (20%, 2 turns)"
- Updated tag names for Elemental Vulnerability
- Added Element Boost tags


classes/battle/BattleAttack.php
```diff
@@ -6,6 +6,15 @@
...
+    public float $piercing_percent = 0;
+    public float $substitution_percent = 0;
+    public float $counter_percent = 0;
+    public float $immolate_percent = 0;
+    public float $immolate_raw_damage = 0;
+    public float $recoil_percent = 0;
+    public float $recoil_raw_damage = 0;
+    public float $countered_percent = 0; // opponent's counter
+    public float $countered_raw_damage = 0; // damage dealt from opponent
+    public string $countered_jutsu_type; // jutsu type of opponent's counter
```
levimeahan, line 19: This is great, nice to keep everything organized with the attack object itself 🎉

classes/battle/BattleManager.php
```diff
@@ -880,8 +913,28 @@ protected function applyAttack(BattleAttack $attack, Fighter $user, Fighter $tar
                                 </p>"; }
                     }
 
+        if ($attack->recoil_percent > 0) {
+            if ($recoil_damage_resisted > 0) {
+                $text .= "<span>-" . $user->getName() . " takes <span class=\"battle_text_{$attack->jutsu->jutsu_type}\">" . round($recoil_damage, 0) . "</span> recoil damage- (resists " . "<span class=\"battle_text_{$attack->jutsu->jutsu_type}\">" . round($recoil_damage_resisted) . "</span>" . " recoil damage)" . '</span></br>';
```
levimeahan, line 918: IMO especially if resist is a % now, we might be able to leave the resist part of this display off and just show the final recoil daamge

I think the absolute number was much more important when resist was not a % and it was hard to tell how much you would really resist

classes/battle/BattleManager.php
```diff
@@ -723,7 +725,31 @@ protected function setupFighterAttack(Fighter $fighter, LegacyFighterAction $act
...
+                case 'immolate':
+                    $attack->immolate_percent += $effect->effect_amount / 100;
+                    $attack->immolate_raw_damage += $this->effects->processImmolate($attack, $target);
+                    break;
+                default:
+                    break;
+            }
+        }
+
+        $attack->raw_damage = $fighter->calcDamage(attack: $attack->jutsu, disable_randomness: $disable_randomness, immolate_raw_damage: $attack->immolate_raw_damage * $attack->immolate_percent);
```
levimeahan, line 752:
```suggestion
        $attack->raw_damage = $fighter->calcDamage(attack: $attack->jutsu, disable_randomness: $disable_randomness, immolate_raw_damage: $attack->immolate_raw_damage * $attack->immolate_percent);
```
Wait, won't this cause immolate to double-dip? It gets added to the attack's raw damage here, and then also triggers separately in `applyAttack`

I think the `applyAttack` is the one we should keep and this is the one we should drop, unlike recoil immolate is not making your attack itself more powerful, it triggers a secondary burst of damage from bleeding/fire/etc that is already on the target

levimeahan (approved):
Immo thing should be fixed but everything else is optional, LFGGGGGG!

admin/entity_constraints.php
```diff
@@ -233,9 +233,51 @@
                 'options' => [Jutsu::USE_TYPE_MELEE, Jutsu::USE_TYPE_PROJECTILE, Jutsu::USE_TYPE_BUFF, Jutsu::USE_TYPE_BARRIER],
             ],
             'effect' => [
-                'data_type' => 'string',
-                'input_type' => 'text',
-                'options' => ['none', 'residual_damage', 'taijutsu_boost', 'ninjutsu_boost', 'genjutsu_boost', 'speed_boost', 'cast_speed_boost', 'speed_nerf', 'weakness_fire', 'weakness_wind', 'weakness_lighting', 'weakness_earth', 'weakness_water', 'evasion_boost', 'evasion_nerf', 'vulnerability', 'offense_nerf', 'resist_boost', 'piercing', 'substitution', 'counter', 'immolate'],
+                    'data_type' => 'string',
+                    'input_type' => 'text',
+                    'options' => [
```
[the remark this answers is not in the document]
aumu1031: Yeaaaah agreed.

PR #509, opened 2023-12-09
Combat v1.2 updates 4

aumu1031:
- Add "Reflect" tag, essentially a Counter that turns incoming damage into a residual
    - Will lean into the Genjutsu playstyle of using residuals as a resource to empower other jutsu
- Add "Delayed Residual" tag, effectively Piercing but for jutsu with lower base power (e.g. Lightning Genjutsu)
- Reworked Immolate to ONLY convert residuals into power for the jutsu
    - Keeps Genjutsu more interactive so both players can play around Immolate effects
    - Lets us use Immolate as more than a combo finisher, residuals become a resource
- Capped Counter/Reflect damage to that of the counter
    - e.g. a Counter that deals 30k direct damage can only deal 30k additional counter damage (in terms of raw damage)
- Fixed elemental tags
- Fixed bug that could cause players to be stuck in combat

Idea behind these changes is so that Gen is more interactive both ways while still focusing on residuals. 

Current playstyle:
- Debuff
- Long Residual
- Medium Residual
- Vulnerability
- Immolate

This is functional but battles become a damage race with little variations in fights.

New sample playstyle:
- Medium Residual (deal damage and prepare to empower other jutsu)
- Reflect (reduce an opponent's early combo and set yourself up)
- Immolate Residual (combo extender, uses previous two residuals to create a stronger one)
- Vulnerability (set up for burst turn)
- Delayed Residual (opponent predicts the burst and uses a Counter, predict this and bypass)

This creates opportunities on both sides for players to interact, less predetermined outcomes.
Hopefully Genjutsu will be more comparable to Nin/Tai while retaining its own identity.

### DIFF
--- a/admin/constraints/jutsu.php
+++ b/admin/constraints/jutsu.php
@@ -42,6 +42,8 @@ $jutsu_effects = [
     'counter',
     'immolate',
     'recoil',
+    'delayed_residual',
+    'reflect',
 ];
 
 $ranks = [];
--- a/classes/battle/BattleEffectsManager.php
+++ b/classes/battle/BattleEffectsManager.php
@@ -94,11 +94,11 @@ class BattleEffectsManager {
             case 'lightning_boost':
             case 'earth_boost':
             case 'water_boost':
-            case 'fire_weakness':
-            case 'wind_weakness':
-            case 'lightning_weakness':
-            case 'earth_weakness':
-            case 'water_weakness':
+            case 'fire_vulnerability':
+            case 'wind_vulnerability':
+            case 'lightning_vulnerability':
+            case 'earth_vulnerability':
+            case 'water_vulnerability':
                 // No changes needed to base number, calculated in applyPassiveEffects
                 break;
             case 'intelligence_boost':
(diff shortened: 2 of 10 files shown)


classes/battle/BattleEffectsManager.php
```diff
@@ -423,7 +428,7 @@ public function applyActiveEffect(Fighter $target, Fighter $attacker, BattleEffe
             return false;
         }
 
-        if($effect->effect == 'residual_damage' || $effect->effect == 'bleed') {
+        if($effect->effect == 'residual_damage' || $effect->effect == 'bleed' || $effect->effect == 'delayed_residual' || $effect->effect == 'reflect_damage') {
```
levimeahan, line 431: This looks like it might be time to make a "damage over time effects" constant array and check here

admin/constraints/jutsu.php
```diff
@@ -42,6 +42,8 @@
     'counter',
     'immolate',
     'recoil',
+    'delayed_residual',
```
levimeahan, line 45: How does this effect behave differently from standard residual? I looked through the code in the rest of the files but don't see any mechanical changes

levimeahan (approved):
Okay I think we should change this a bit but will approve to unblock testing

classes/battle/BattleEffectsManager.php
aumu1031: Makes sense especially since we use this multiple times.

PR #511, opened 2023-12-16
Combat v1.2 updates 5

aumu1031:
image
image

### DIFF
--- a/classes/battle/BattleManager.php
+++ b/classes/battle/BattleManager.php
@@ -1560,4 +1560,8 @@ class BattleManager {
     public function getPatrolId(): int {
         return $this->battle->patrol_id;
     }
+
+    public function getEffects(): BattleEffectsManager {
+        return $this->effects;
+    }
 }
--- a/classes/battle/BattleManager.php
+++ b/classes/battle/BattleManager.php
@@ -754,7 +754,7 @@ class BattleManager {
                     break;
                 case 'reflect':
                     $attack->reflect_percent += $effect->effect_amount / 100;
-                    $attack->reflect_duration = $effect->effect_length;
+                    $attack->reflect_duration = min($effect->effect_length, 1);
                     break;
                 default:
                     break;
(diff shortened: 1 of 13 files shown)


classes/battle/BattleManager.php
```diff
@@ -1185,7 +1184,7 @@ public function jutsuCollision(
         $player1_evasion_stat_amount = $this->getEvasionPercent($player1, $player1_jutsu, $player2->getBaseStatTotal());
         $player2_evasion_stat_amount = $this->getEvasionPercent($player2, $player2_jutsu, $player1->getBaseStatTotal());
 
-        if($player1_evasion_stat_amount >= $player2_evasion_stat_amount && $player2_jutsu_is_attack) {
+        if($player1_evasion_stat_amount >= $player2_evasion_stat_amount) {
```
levimeahan, line 1187: Actually curious about this one, why apply evasion against e.g. barriers? That would seem to make speed builds stronger than expected because they reduce the power of their opponent's barriers

classes/battle/BattleManager.php
```diff
@@ -1561,4 +1560,8 @@ protected function chooseAndSetAIAction(Fighter $ai) {
     public function getPatrolId(): int {
         return $this->battle->patrol_id;
     }
+
+    public function getEffects(): BattleEffectsManager {
+        return $this->effects;
+    }
```
levimeahan, line 1566:
```suggestion
    /**
      * @return BattleEffect[]
      */
    public function getActiveEffects(): array {
        return $this->effects->active_effects;
    }
```
If possible let's do this instead so the full options of the effects manager are not exposed (this is to force effects logic into this class or an explicit dependency)

levimeahan:
image
Feels a little weird to have a default jutsu buried in the middle of the list, could we move those to the top/start or bottom/end?

levimeahan:
This UI is awesome 🙌 So much nicer

levimeahan (approved):

classes/battle/BattleManager.php
aumu1031: Normalizes barrier strength across different builds so that someone with 50%/50% BL/Off skill doesn't have barriers almost twice as strong as someone with 25%/25%/50% BL/Off/CSpd, even though normally both players do the same damage to each other with other jutsu.

classes/battle/BattleManager.php
aumu1031: Makes sense.

classes/battle/BattleManager.php
levimeahan: Ahh, I see. This seems pretty unintuitive from a "how does combat work" perspective, but I can see how it's simpler to balance this way

PR #512, opened 2023-12-16
Weekend event

aumu1031:
- Add +50% bonus XP on weekends if no other event active
- Jutsu training affected by bonus XP

### DIFF
--- a/classes/System.php
+++ b/classes/System.php
@@ -10,6 +10,7 @@ require_once __DIR__ . '/Router.php';
 require_once __DIR__ . '/Route.php';
 require_once __DIR__ . '/../classes/event/DoubleExpEvent.php';
 require_once __DIR__ . '/../classes/event/DoubleReputationEvent.php';
+require_once __DIR__ . '/../classes/event/BonusExpWeekend.php';
 
 /*	Class:		System
 	Purpose: 	Handle database connection and queries. Handle storing and printing of error messages.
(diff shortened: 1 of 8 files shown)


classes/System.php
```diff
@@ -699,6 +700,12 @@ public function checkForActiveEvent(): void {
         if($this->testNotifications['event'] && is_null($this->event) && $this->isDevEnvironment()) {
             $this->event = new DoubleExpEvent($current_datetime->modify("+2 weeks"));
         }
+
+        // If no other event active on weekend, activate bonus XP
+        if (!isset($this->event) && (date('w') == 0 || date('w') == 6)) {
```
levimeahan, line 705:
```suggestion
        if (!isset($this->event) && (System::currentDayOfWeek() == 0 || System::currentDayOfWeek() == 6)) {
```

Can you add a helper for this (we have the others for currentDay/CurrentMonth/etc already):

```
    /**
     * @return int day of the week: 0 (Sunday) - 6 (Saturday)
     */
    public static function currentDayOfWeek(): int {
        return (int) date('w');
    }
```

classes/event/Event.php
```diff
@@ -3,4 +3,5 @@
 abstract class Event {
     public string $name;
     public DateTimeImmutable $end_time;
+    public float $exp_modifier = 1;
```
levimeahan, line 6:
```suggestion
    public float $exp_gain_multiplier = 1;
```

levimeahan (approved):

classes/System.php
aumu1031: Sure!

PR #515, opened 2023-12-16
version update

aumu1031:


classes/System.php
```diff
@@ -107,8 +107,8 @@ class System {
 
     // Default layout
     const DEFAULT_LAYOUT = 'new_geisha';
-    const VERSION_NUMBER = '0.10.0';
-    const VERSION_NAME = '0.10 Warring Shadows';
+    const VERSION_NUMBER = '0.10.1';
+    const VERSION_NAME = '0.10.1 Warring Shadows';
```
levimeahan, line 111:
```suggestion
    const VERSION_NUMBER = '0.11';
    const VERSION_NAME = '0.11 Warring Shadows';
```
This combat update deserves a full number bump, I think (but we could leave title as Warring Shadows if desired)

levimeahan (approved):

PR #517, opened 2023-12-17
battle qol

aumu1031:
- new button at top (same bar as refresh battle) jumps you straight to the jutsu select
- double clicking a jutsu uses the jutsu
- fixed weirdness with jutsu text overflowing
- clicking taijutsu adds the overlay thing to the hand seal rather than making the page jump up/down
- hand seal overlay now uses the same color as the container rather than blinding white

image

### DIFF
--- a/templates/battle/action_prompt.php
+++ b/templates/battle/action_prompt.php
@@ -188,6 +188,9 @@ $prefill_item_id = $_POST['item_id'] ?? '';
         var currentlySelectedJutsu = false;
         var lastJutsu, firstJutsu = false;
         $('.jutsuName').click(function(){
+            if (lastJutsu == this) {
+                document.getElementById('submitbtn').click();
+            }
 
             if(lastJutsu !== this && firstJutsu) {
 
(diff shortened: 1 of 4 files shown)


templates/battle/action_prompt.php
```diff
@@ -462,13 +469,6 @@ function handleKeyUp(event) {
                             <div class="jutsu_details">
                                 <div class="jutsu_details_row">
                                     <span class="jutsu_details_power">Power: <?= round($attack->power, 2) ?></span>
-                                    <span class="jutsu_details_cooldown">
-                                        <?php if ($cd_left > 0): ?>
-                                                CD: (<?= $cd_left ?>) <?= $attack->cooldown ?> turns
-                                        <?php else: ?>
-                                                CD: <?= $attack->cooldown ?> turns
-                                        <?php endif; ?>
-                                    </span>
```
levimeahan, line 471: Hmm, why do we need to remove this?

levimeahan (approved):

PR #528, opened 2023-12-22
Admin panel additions

HitoriWoB:
Removed access to user admin tools for Content Admins
Added temporary method to reset user passwords without needing direct access to DB

### DIFF
--- a/classes/StaffManager.php
+++ b/classes/StaffManager.php
@@ -672,23 +672,23 @@ class StaffManager {
     }
 
     /** ADMIN PANEL METHODS **/
-    public function getAdminPanelPerms($type): array {
+    public function getAdminPanelPerms(string $type, bool $permission_check = false): array {
         switch($type) {
             // Keep create content and edit content in line with each other
             case 'create_content':
-                if($this->isContentAdmin()) {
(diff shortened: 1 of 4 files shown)


classes/StaffManager.php
```diff
@@ -672,23 +672,23 @@ public function getBanLengths() {
     }
 
     /** ADMIN PANEL METHODS **/
-    public function getAdminPanelPerms($type): array {
+    public function getAdminPanelPerms(string $type, bool $permission_check = false): array {
```
levimeahan, line 675: What is the use case for not checking permissions here?

levimeahan (approved):

classes/StaffManager.php
HitoriWoB: Temp fix until I overhaul admin panel. Prior to this, the panel links would display properly.

However on permission check the return would be an empty array (expected) and then just skip this permissions statement. This allowed content admins access to all tools if the link was known. 

This fix tweaks that a bit to actually return the arrays and do a proper permissions check. I'll get it resolved properly post-character launch with the admin panel fixes I want to get done.

PR #530, opened 2023-12-23
100% pre-heal

aumu1031:
With the state of alt-sniping, war mechanics, and the ability to use healing items immediately after battle, pre-heal 100% creates an even playing field. Especially considering battles are generally evenly matched now rather than before where you would need lower pre-heal just to be evenly matched against your counter.

Personal suggestion: pre-heal that drains pools to use versus consumables so you still feel like you're making progress when attacking someone who is weakened, even if that battle doesn't result in a win.

### DIFF
--- a/classes/battle/Battle.php
+++ b/classes/battle/Battle.php
@@ -18,7 +18,7 @@ class Battle {
     const MAX_TURN_LENGTH = 40;
     const PREP_LENGTH = 20;
 
-    const MAX_PRE_FIGHT_HEAL_PERCENT = 85;
+    const MAX_PRE_FIGHT_HEAL_PERCENT = 100;
 
     const TEAM1 = 'T1';
     const TEAM2 = 'T2';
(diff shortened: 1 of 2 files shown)


levimeahan:
> With the state of alt-sniping, war mechanics, and the ability to use healing items immediately after battle, pre-heal 100% creates an even playing field. Especially considering battles are generally evenly matched now rather than before where you would need lower pre-heal just to be evenly matched against your counter.
>
>Personal suggestion: pre-heal that drains pools to use versus consumables so you still feel like you're making progress when attacking someone who is weakened, even if that battle doesn't result in a win.
I agree both that network latency should not give an advantage, and that 100% pre-heal should not be given without some kind of limiter to prevent extended win streaks. I think before we merge this we should agree on what the right limiter is and include that in this change

> pre-heal that drains pools to use versus consumables
This kind of makes sense but also I feel might be stepping on medical squad territory.

Hmm... Let's discuss more maybe on discord about what the right option is here

PR #533, opened 2023-12-25
Handle Travel Exceptions

aumu1031:
Use existing travel message logic to display "normal" exceptions rather than flooding the error log with uncaught exceptions.

image


levimeahan (approved):
🙏 nice!

PR #541, opened 2023-12-31
New year boost

HitoriWoB:


aumu1031:
Since the class is being changed to a general constructor for an EXP event rather than just the weekends, its more clear if you rename the class from `Bonus EXP Weekend` to `Bonus EXP Event`.

HitoriWoB:
> Since the class is being changed to a general constructor for an EXP event rather than just the weekends, its more clear if you rename the class from `Bonus EXP Weekend` to `Bonus EXP Event`.
> 
> But its a little weird if the default value is `Bonus EXP Weekend` in that case as well, so it makes sense to change that to `Bonus EXP Event` and pass the specific name for instances where its auto-set on weekends.

Oh, edited the wrong file. Let me fix that and modify the DoubleExpEvent.

levimeahan (approved):

PR #543, opened 2024-01-01
Kage Record + Sumu Updates

aumu1031:
image

### DIFF
--- a/classes/village/VillageManager.php
+++ b/classes/village/VillageManager.php
@@ -2118,7 +2118,7 @@ class VillageManager {
         switch ($proposal_type) {
             case self::PROPOSAL_TYPE_BREAK_ALLIANCE:
                 $notification_type = NotificationManager::NOTIFICATION_DIPLOMACY_END_ALLIANCE;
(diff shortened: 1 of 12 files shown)


classes/village/VillageManager.php
```diff
@@ -373,20 +373,24 @@ public static function getVillageSeats(System $system, int $village_id): array {
...
+                        $provisional_days_remaining = ": " . System::timeRemaining(time() - $seat['seat_start'], format: 'days');
+                    }
                     $seats[] = new VillageSeatDto(
                         seat_key: 'kage',
                         seat_id: $seat['seat_id'],
                         user_id: $seat['user_id'],
                         village_id: $seat['village_id'],
                         seat_type: $seat['seat_type'],
-                        seat_title: $seat['seat_title'],
+                        seat_title: $seat['seat_title'] . $provisional_days_remaining,
```
levimeahan, line 393: Personally not sure about making this so generic (everywhere we show the title we show the # of days), what if we added it specifically as a display after the title on the Village HQ?

levimeahan (approved):

classes/village/VillageManager.php
aumu1031: SGTM, making the tweak

PR #548, opened 2024-01-01
Homepage messages & server maintenance

HitoriWoB:
- Fixed homepage maintenance mode message
- Patched reputation reset timer bug on day of reset
- Added support for auto-opening and closing based on expected start & end time
   - Support for manual closing for an unknown time frame is still support through system vars

image
image

### DIFF
--- a/classes/System.php
+++ b/classes/System.php
@@ -68,6 +68,7 @@ class System {
     public bool $enable_dev_only_features;
 
     public bool $SC_OPEN;
+    public ?DateTimeImmutable $UPDATE_MAINTENANCE;
     public bool $register_open;
     public bool $USE_NEW_BATTLES = false;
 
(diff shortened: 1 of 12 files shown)


classes/System.php
```diff
@@ -727,6 +731,52 @@ public function checkForActiveEvent(): void {
...
+        }
+
+
+        return null;
+    }
+
+    public function checkForMaintenanceEnd(): string {
+        if(!$this->SC_OPEN && $this->UPDATE_MAINTENANCE) {
+            $mins = ceil(($this->UPDATE_MAINTENANCE->getTimestamp() - time()) / 60);
+            $mins += 5 - ($mins % 5);
```
levimeahan, line 761: Can you add a comment about what this is doing, was a little unclear at first 🙏

levimeahan (approved):
This is cool all_of_the

LGTM with one non-blocking comment

PR #551, opened 2024-01-02
Battle Details Toggle

aumu1031:
image


HitoriWoB (approved):
Could consider adding independent toggles for just battle details or just last turn details.
Adds a bit of complexity to your "show/hide all".

aumu1031:
> Could consider adding independent toggles for just battle details or just last turn details. Adds a bit of complexity to your "show/hide all".

Yep, let's see what complaints/suggestions people have for this since it was requested.

PR #552, opened 2024-01-03
Bloodline Jutsu Display

aumu1031:
- Display bloodline jutsu in rank order
- Fix issue where bloodline jutsu would show the actual power of the residual damage after use instead of the % (e.g. 478437 instead of 25%)

### DIFF
--- a/templates/battle/action_prompt.php
+++ b/templates/battle/action_prompt.php
@@ -475,10 +475,10 @@ $prefill_item_id = $_POST['item_id'] ?? '';
                                             <?php
                                             if ($effect->effect != "none") {
                                                 echo in_array($effect->effect, ["substitution", "counter", "piercing", "recoil"])
-                                                    ? System::unSlug($effect->effect) . " (" . round($effect->effect_amount, 0) . "%)"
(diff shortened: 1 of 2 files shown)


templates/battle/action_prompt.php
```diff
@@ -475,10 +475,10 @@ function handleKeyUp(event) {
                                             <?php
                                             if ($effect->effect != "none") {
                                                 echo in_array($effect->effect, ["substitution", "counter", "piercing", "recoil"])
-                                                    ? System::unSlug($effect->effect) . " (" . round($effect->effect_amount, 0) . "%)"
+                                                    ? System::unSlug($effect->effect) . " (" . round($effect->display_effect_amount, 0) . "%)"
```
HitoriWoB, line 478: Not blocking, but could consider making this a method for easier maintenance.
`$effect->getDisplay()`

HitoriWoB (approved):

templates/battle/action_prompt.php
aumu1031: SGTM

PR #554, opened 2024-01-03
CSS Reorganization

aumu1031:
- Split common structure CSS out of New Geisha / Sumu
- Move old overrides into separate CSS files for New Geisha / Sumu
- Sort remaining styles in main layout files


style/new_geisha/new_geisha_legacy.css
```diff
@@ -0,0 +1,611 @@
+/* Helper Classes */
```
levimeahan, line 1: What's the role of this file vs `new_geisha.css`?

layout/new_geisha.php
```diff
@@ -31,7 +31,9 @@ function getNewGeishaLayout(System $system, bool $enable_mobile_layout): Layout
 	<title>Shinobi Chronicles RPG</title>
 
 	<link rel='stylesheet' type='text/css' href='{$system->getCssFileLink("style/geisha/geisha.css")}' />
```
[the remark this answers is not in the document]
aumu1031: That is the next step and something I would prefer in this PR, but is also really tedious!
At the moment I have removing `geisha.css` about ~half finished locally but I'm thinking we make that a new branch after merging this PR since its blocking the creation of new pages (Jutsu Page, Manual).

style/new_geisha/new_geisha_legacy.css
aumu1031: It has all of the existing styles that go against the new design. So eventually these when we update the CSS of existing pages (like new Profile, new Travel, new VillageHQ) we'll be able to totally delete `new_geisha_legacy.css`.

So its basically a way to say, everything in `new_geisha.css` is good, everything in `new_geisha_legacy.css` needs replaced.

levimeahan (approved):

style/new_geisha/new_geisha_legacy.css
levimeahan: Makes sense!

PR #560, opened 2024-01-07
Optimize Chat

aumu1031:


classes/chat/ChatManager.php
```diff
@@ -78,33 +78,75 @@ private function fetchPosts(?int $starting_post_id = null, int $max_posts = self
...
+            }
+            $user_data_by_id[$row['user_id']] = $row;
+        }
         $posts = [];
-        while($row = $this->system->db->fetch($result)) {
+        foreach ($result as $row) {
             $post = ChatPostDto::fromDb($row);
 
             //Base data
             $post->avatar = './images/default_avatar.png';
```
HitoriWoB, line 141: Just a note for us, could move defaults like these into the ChatPostDto.

classes/chat/ChatManager.php
```diff
@@ -78,33 +78,75 @@ private function fetchPosts(?int $starting_post_id = null, int $max_posts = self
...
+                        $user_data_result['avatar_style'] = "avy_round";
+                    }
+                    // if custom setting is not set
+                    if (!isset($user_data_result['avatar_frame'])) {
+                        $user_data_result['avatar_frame'] = "avy_frame_default";
+                    }
+                    $user_data_by_name[$row['user_name']] = $user_data_result;
+                }
+            }
+        }
```
HitoriWoB, line 115: Doesn't appear that this handles the situation where there is no UID and user isn't found by name.

classes/chat/ChatManager.php
aumu1031: Good point, taking note as well

classes/chat/ChatManager.php
aumu1031: There was a check for if user data is found for a post since this case can currently happens as well (when users change their name). Updated it slightly here to check if data was found either by ID or name:
```            //Fetch user data
            $user_data = null;
            if (isset($user_data_by_id[$post->user_id])) {
                $user_data = $user_data_by_id[$post->user_id];
            }
            else if (isset($user_data_by_name[$post->user_name])) {
                $user_data = $user_data_by_name[$post->user_name];
            }```

levimeahan (approved):

PR #562, opened 2024-01-08
Pre-Heal Update

aumu1031:
- Pre-Heal to 95%
- Healing items are now a %
- Cap out of combat healing to pre-heal limit
- Added configuration for per-item carry limit
- Reworked router logic so we can configure conditional overrides for particular pages that are otherwise set to "only in village".

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -134,7 +134,8 @@ return $routes = [
         function_name: 'healingShop',
         menu: Route::MENU_ACTIVITY,
         battle_ok: false,
-        village_ok: Route::VILLAGE_OR_COLOSSEUM,
+        village_ok: Route::ONLY_IN_VILLAGE,
+        village_restriction_overrides: [Route::COLOSSEUM_OVERRIDE => "true", Route::TOWN_OVERRIDE => "true", Route::CASTLE_OVERRIDE => "true"],
     ),
     26 => new Route(
         file_name: 'viewBattles.php',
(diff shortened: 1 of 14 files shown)


config/routes.php
```diff
@@ -134,7 +134,8 @@ function_name: 'spar',
         function_name: 'healingShop',
         menu: Route::MENU_ACTIVITY,
         battle_ok: false,
-        village_ok: Route::VILLAGE_OR_COLOSSEUM,
+        village_ok: Route::ONLY_IN_VILLAGE,
+        village_restriction_overrides: [Route::COLOSSEUM_OVERRIDE => "true"],
```
levimeahan, line 138: I think this is the point we should move away from `village_ok` entirely to a more granular `allowed_location_types` array that maps from location type => allowed or not

```suggestion
        allowed_location_types: [
            Route::LOCATION_VILLAGE,
            Route::LOCATION_COLOSSEUM
        ],
```
If this param is not provided, we could assume all location types are okay. But if provided, we only let the route trigger if the current location type matches one in this array

levimeahan (approved):
Approving to unblock, but please address the route location type feedback before merging 👍

config/routes.php
aumu1031: I like `allowed_location-types` solution much better, building on `village_ok` logic makes this overly complicated.
I will rework this and apply to all existing routes in a follow up.

levimeahan (approved):
Awesome, this is much better!

PR #564, opened 2024-01-08
Battle Rounds

aumu1031:
- Adds optional parameter to battle initialization
- Battles have rounds, tracks winner of each round and resets to clean state at start of next
- Winner declared if sufficient wins reached (e.g. 2 wins in a 3 round battle will skip the 3rd round)
- If tied by end of round limit, will add additional rounds until the next win
- Challenges are now set to 3 rounds
- Reputation bonus for challenges from 15% -> 5%

### DIFF
--- a/classes/battle/BattleManager.php
+++ b/classes/battle/BattleManager.php
@@ -666,23 +666,25 @@ class BattleManager {
             return Battle::DRAW;
         }
         // if team1 majority of wins (total rounds)
-        if ($this->battle->team1_wins > floor($this->battle->rounds / 2) || ($this->battle->current_round > $this->battle->rounds && $this->battle->team1_wins > $this->battle->team2_wins)) {
-             return Battle::TEAM1;
+        if ($this->battle->team1_wins > floor($this->battle->rounds / 2) || ($this->battle->round_count >= $this->battle->rounds && $this->battle->team1_wins > $this->battle->team2_wins)) {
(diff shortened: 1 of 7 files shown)


classes/battle/BattleManager.php
```diff
@@ -651,6 +651,66 @@ private function checkForWinner(): string {
...
+            $this->battle->player1_last_damage_taken = 0;
+            $this->battle->player2->health = $this->battle->player2->max_health;
+            $this->battle->fighter_health[$this->battle->player2->combat_id] = $this->battle->player2->health;
+            $this->battle->player2_last_damage_taken = 0;
+            $this->effects->active_effects = [];
+            $this->battle->jutsu_cooldowns = [];
+            $this->battle->turn_count = 0;
+            $this->battle->turn_time = time();
+            $this->battle->player1_time = Battle::MAX_TURN_LENGTH;
+            $this->battle->player2_time = Battle::MAX_TURN_LENGTH;
```
levimeahan, line 690: This looks like the same logic as below, could you extract it to a function?

levimeahan (approved):

classes/battle/BattleManager.php
aumu1031: Yeah both cases there are completely identical (for now) - I'll copy this into a helper function (`resetBattle()`) and we can define differences with how extra rounds work separately.

PR #571, opened 2024-01-12
Monthly Reset

aumu1031:


levimeahan (approved):
One thing to change with this one - Can we add a `prev_month_points` column to easily show leaderboards of the last month?

aumu1031:
> One thing to change with this one - Can we add a `prev_month_points` column to easily show leaderboards of the last month?

Done!

PR #572, opened 2024-01-12
Exam Typo

aumu1031:


templates/sensei_exam.php
```diff
@@ -42,7 +42,7 @@
                 <div><input type="radio" name="question1" value="1c" /> <label>Jutsu Levels</label></div>
                 <div><input type="radio" name="question1" value="1d" /> <label>All of the above</label></div>
                 <div><input type="radio" name="question1" value="1e" /> <label>None of the above (pure chance)</label></div>
-                <p class="question_label">Which of these Forbidden Seal benefits is not granted by Twin Dragon Seal?</p>
+                <p class="question_label">Which of these Forbidden Seal benefits is not granted by Twin Sparrow Seal?</p>
```
levimeahan, line 45: We could go one step further:
```suggestion
                <p class="question_label">Which of these Forbidden Seal benefits is not granted by <?= ForbiddenSeal::$forbidden_seal_names[1] ?>?</p>
```

levimeahan (approved):

templates/sensei_exam.php
aumu1031: Big brain move, let's do it.

PR #575, opened 2024-01-14
Global message fix

HitoriWoB:


HitoriWoB:
Removed server closure from mod panel, to remove requirement of posing globals.
Only head admin can begin maint/close server.
Managed through admin panel.

levimeahan (approved):

PR #579, opened 2024-01-14
Battle Log / Round Compatibility and Tweaks

aumu1031:
- Use 2D array for fighter health
- Update columns, use SMALLINT(6) and default value of 1 for `round_number`
- Add `round_number` to `battle_logs` and update indexes
- Update battle history display for rounds
- Fix warnings on undefined avatar size

### DIFF
--- a/pages/viewBattles.php
+++ b/pages/viewBattles.php
@@ -311,13 +311,14 @@ function viewBattles() {
                 }
                 // get battle
                 $battle_result = $system->db->query("
-                    SELECT `player1`, `player2`, `battle_background_link` FROM `battles` WHERE `battle_id` = {$battle_id}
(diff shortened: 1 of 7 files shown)


pages/viewBattles.php
```diff
@@ -295,7 +295,8 @@ function viewBattles() {
 
 
         /* View Log */
-        $logs_result;
+        $p1_avatar_size = 200;
+        $p2_avatar_size = 200;
```
HitoriWoB, line 299: Assuming this is intentional for new battle UI, since 200 is premium size?

HitoriWoB (approved):
Looks good, assuming the premium avatar size is intended as a default for new UI.

pages/viewBattles.php
aumu1031: Yep just defaults for error handling, the p1 shouldn't be used.

PR #580, opened 2024-01-15
Chuunin Elder

aumu1031:


HitoriWoB (approved):
With the lack of elders currently, I feel this is fine. If the seats start being held by Jonin more easily, we should revert this back to the current requirements.

PR #584, opened 2024-01-16
Characters Prep Phase 1

HitoriWoB:
- Index cleanup
 - Change user verification to a soft verification. New users will need to validate email once email service is restored


HitoriWoB:
System refactor in progress. Should have this done tonight.

secure/vars.sample.php
```diff
@@ -5,14 +5,19 @@
...
 
 //Host needs to match the DB's host. For docker setup, rename it to "shinobi-chronicles-db".
-$host = 'localhost';
-$username = 'database_username';
-$password = 'database_password';
-$database = 'database_name';
+$HOST = 'localhost';
+$USERNAME = 'database_username';
+$PASSWORD = 'database_password';
+$DATABASE = 'database_name';
```
levimeahan, line 23: Will this break the existing usages of `System` if we merge as is?

db/seeds/BaseStructureAndData.php
```diff
@@ -932,6 +932,11 @@ public function run(): void {
             (5, 'Mist', '16.10', 0, 0);"
         );
 
+        // System Storage
+    $this->execute(
+        "INSERT INTO `system_storage` (`global_message`, `time`) VALUES ('', '" . time() . "');"
+    );
```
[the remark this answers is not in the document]
HitoriWoB: Intended to for future seeds. I think most people running local environments won't really notice this, since they're not using global messages/maintenance.

secure/vars.sample.php
HitoriWoB: Yes, right now things are in a bit of a broken state. I'm currently working through some bugs with SysV2. I could revert some of these changes if we're wanting to push out just the index changes. I'd think I'll have things running smoothly within a few hours of getting home tonight.

db/seeds/BaseStructureAndData.php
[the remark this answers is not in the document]
levimeahan: What would happen if we didn't have this line?

secure/vars.sample.php
levimeahan: Can wait for you to have this ready to go, just making sure this is something we're fixing 👍

db/seeds/BaseStructureAndData.php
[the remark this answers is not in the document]
HitoriWoB: Game breaks if a global is sent out and server maintenance fails.
I suppose it could be added via a migration as a truncate => add row.

index.php
```diff
@@ -483,6 +101,11 @@
         //Ban info
         require 'templates/ban_info.php';
 
+        // Load user record
+        $route = Router::$routes[34];
```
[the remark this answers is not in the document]
HitoriWoB: Will do. Thinking this will likely move to character select/user settings or something, since we can just restrict selecting/changing characters with game bans.

secure/vars.sample.php
HitoriWoB: Will have it fixed 💯 I'll likely have some additional variables that will need added on dev & production.

classes/Auth.php
```diff
@@ -23,10 +23,126 @@ public static function getUserFromSession(System $system): User {
...
+            if(!$system->db->last_num_rows) {
+                throw new RuntimeException("User not found!");
+            }
+            
+            // Fetch data
+            $user_data = $system->db->fetch($result);
+
+            // Block login to unauthorized users during server maint
+            if(!$system->SC_OPEN && !StaffManager::hasServerMaintAccess(staff_level: $user_data['staff_level'])) {
+                throw new RuntimeException("Server is closed! Try againsoon.");
```
levimeahan, line 56:
```suggestion
                throw new RuntimeException("Server is closed! Try again soon.");
```

classes/_sysRequires.php
```diff
@@ -0,0 +1,18 @@
...
+require_once __DIR__ . '/MarkdownParser.php';
+require_once __DIR__ . '/API.php';
+require_once __DIR__ . '/Layout.php';
+require_once __DIR__ . '/Router.php';
+require_once __DIR__ . '/Route.php';
+
+require_once __DIR__ . '/../classes/event/DoubleExpEvent.php';
+require_once __DIR__ . '/../classes/event/DoubleReputationEvent.php';
+require_once __DIR__ . '/../classes/event/BonusExpWeekend.php';
+require_once __DIR__ . '/../classes/event/HolidayBonusEvent.php';
```
levimeahan: Let's put this back into the System.php class, having all the requires at the top of the file is not a bad thing

classes/Route.php
```diff
@@ -28,5 +28,12 @@ public function __construct(
 
         public ?Closure $user_check = null,
         public bool $dev_only = false,
+
+        // Layout controls
+        public bool $render_header = true,
+        public bool $render_sidebar = true,
+        public bool $render_topbar = true,
+        public bool $render_content = true,
+        public bool $render_footer = true
```
levimeahan, line 37: This does not seem like something that's really route-specific. Homepage is a special case but unless we have multiple cases, let's not add these as extra route parameters, it's okay for the homepage to get hardcoded special treatment

db/seeds/BaseStructureAndData.php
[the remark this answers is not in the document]
levimeahan: Ah, I see. Hmm... Okay, I think we can say this is a special case since we manually applied this already and everyone that's already set up a dev environment would have had to manually set it up too. Let's leave it in the seed file 👍

index.php
```diff
@@ -1,477 +1,102 @@
...
-    ini_set('display_errors', 'On');
+    ini_set(option: 'display_errors', value: 'On');
+}
+// Disable errors on production for non-developers
+else {
+    $value = 'Off';
+    if(isset($_SESSION['user_id']) && in_array($_SESSION['user_id'], System::$developers)) {
+        $value = 'On';
+    }
+    ini_set(option: 'display_errors', value: $value);
```
levimeahan, line 20: It's okay to assume error display is off by default in prod (that's the way we want to set it for security reasons in config, not at runtime in this file)
```suggestion
    if(isset($_SESSION['user_id']) && in_array($_SESSION['user_id'], System::$developers)) {
        ini_set(option: 'display_errors', value: 'On');
    }
```

index.php
```diff
@@ -1,477 +1,102 @@
...
-            $verification_code = sha1(mt_rand(1, 1337000));
-
-            $village_coords = new TravelCoords($villages[$village]['x'], $villages[$village]['y'], $villages[$village]['map_id']);
+    $route = Router::$routes['home'];
+    $system->layout->renderBeforeContentHTML(
+        system: $system,
+        player: null, // No need to check here, no session set
+        page_title: $route->title,
+        render_header: $route->render_header, render_sidebar: $route->render_sidebar,
+        render_topbar: $route->render_topbar, render_content: $route->render_content
```
levimeahan, line 36: Would `$system->layout->renderStaticPageHeader($route->title)` work for you here?

We can do this cleanup after, but I think that or something is a preferable option over having all these parameters that disable a lot of what this function does, for only this case

index.php
```diff
@@ -615,123 +256,30 @@
...
-            $system->printMessage(true);
-        }
-    }
+    // Render after content
+    $PAGE_LOAD_TIME = microtime(as_float: true) - $PAGE_LOAD_START;
+    $system->layout->renderAfterContentHTML(
+        system: $system, player: $player ?? null,
+        page_load_time: $PAGE_LOAD_TIME,
+        render_content: $RENDER_CONTENT,
+        render_hotbar: false
```
levimeahan, line 280: `render_hotbar` is not a parameter on this function AFAIK

support.php
```diff
@@ -16,21 +16,30 @@
...
+    $system->setLayoutByName($player->layout);
     $staff_level = $player->staff_level;
     $user_id = $player->user_id;
 
+    $render_sidebar_header = true;
     $supportSystem = new SupportManager($system, $player);
 }
 else {
-    $layout = $system->setLayoutByName("shadow_ribbon");
+    $render_sidebar_header = false;
```
levimeahan, line 27: What this ends up representing is "is the player logged in, or not". That's information the `renderBeforeContentHTML` function already has (is player null or not), could we tie the logic you're thinking of to that?

levimeahan (approved):

index.php
HitoriWoB: I believe this is tied to the built-in rage bar. Looks like we can drop it because it will only render on dev.

index.php
HitoriWoB: image
I think this is a solid plan, just need to do some tweaks with Layout to make it work later.

index.php
HitoriWoB: Okay, changed that logic. Kept existing logic in place because I wasn't sure if it was needed.

PR #591, opened 2024-01-24
Update Vote Logic

aumu1031:
image


classes/village/VillageManager.php
```diff
@@ -1682,7 +1712,7 @@ public static function enactProposal(System $system, User $player, int $proposal
             $message .= "\n You have gained {$rep_adjustment} Reputation!";
             $player->updateData();
         } else if ($rep_adjustment < 0) {
-            $player->reputation->subtractRep($rep_adjustment, UserReputation::ACTIVITY_TYPE_UNCAPPED);
+            $player->reputation->subtractRep((-1 * $rep_adjustment), UserReputation::ACTIVITY_TYPE_UNCAPPED);
```
HitoriWoB, line 1715: If this has the possibility of being anything but a positive integer, I think using `abs()` would be more appropriate. It could have unintended result in the `subtractRep()` method.
image

aumu1031: Makes sense to me!

HitoriWoB (approved):

PR #594, opened 2024-01-25
Characters prep phase 2

HitoriWoB:
Refactored forbidden seal
Added orange name to T2
Added platinum name to T3

Notable changes to seals:
Tier 1
- Avatars now allow a file size of 1.5MB (up from 1MB)

Tier 2
- Added orange name color
- Max inbox message length 1750 (+25%)

Tier 3
- Added platinum name color
- Avatars now allow a file size of 2.5MB (up from 2MB)
- Inbox size is now 100 (up from 75)
- Max inbox message length 2000 (+50%)
- Max chat post length 450 (up from 400)

image
image


classes/ForbiddenSeal.php
```diff
@@ -1,336 +1,398 @@
...
+        return match($this->level) {
+            0 => "",
+            default => json_encode(array('level' => $this->level, 'time' => $this->seal_end_time))
+        };
+    }
+    public function calcRemainingCredit(): int {
+        match($this->level) {
+            1 => $days_per_ak = 6,
+            2 => $days_per_ak = 2,
+            default => $days_per_ak = 1
```
levimeahan, line 289: Ideally this is something we calculate from the place we define the costs of each tier, as is if we changed the costs in the future this wouldn't automatically be synchronized and there isn't necessarily a way for us to remember we need to change this

classes/ForbiddenSeal.php
```diff
@@ -1,336 +1,398 @@
...
+            $count++;
+            if($count%2 == 0) {
+                $return .= "<br /> ";
+            }
+            else {
+                $return .= '/';
+            }
+        }
+        return substr($return, 0, strlen($return)-1);
+    }
```
levimeahan, line 316: Having display calculated in a class function is a pattern I think we should move away from - Could you move this to the template where it's used? Okay to make it a helper function inside the template

pages/premium_shop.php
```diff
@@ -469,7 +469,11 @@ function premiumShop(): void {
                 if (!isset($_POST['change_forbidden_seal'])) {
                     // Convert remaining premium time to days and calculate AK value
                     $akCredit = $player->forbidden_seal->calcRemainingCredit();
-                    $new_seal = new ForbiddenSeal($system, $seal_level);
+                    $new_seal = ForbiddenSeal::fromDb(
+                        system: $system,
+                        seal_level: $seal_level,
+                        seal_end_time: time() + ForbiddenSeal::SECONDS_IN_DAY
```
levimeahan, line 475: This seems a little problematic that we're putting in a fake value in a lot of different places - Could we make `seal_end_time` nullable? Then we have a clear understanding that if there's no `seal_end_time`, this isn't representing an active player's seal

levimeahan (approved):

PR #598, opened 2024-01-28
Fix Challenge Exploit

aumu1031:
Fix players being able to resign while challenged and let their friends claim the seat before the challenger.

### DIFF
--- a/classes/village/VillageManager.php
+++ b/classes/village/VillageManager.php
@@ -354,6 +354,8 @@ class VillageManager {
         $message = '';
         $time = time();
         $player_seat = $player->village_seat;
+        $player_seat_id = $player_seat->seat_id;
+        $player_seat_type = $player_seat->seat_type;
         // clear active votes
         $system->db->query("DELETE `vote_logs` FROM `vote_logs` INNER JOIN `proposals` on `vote_logs`.`proposal_id` = `proposals`.`proposal_id` WHERE `vote_logs`.`user_id` = {$player->user_id} AND `proposals`.`end_time` IS NULL");
         // exit seat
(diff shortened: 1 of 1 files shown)


classes/village/VillageManager.php
```diff
@@ -347,25 +347,74 @@ public static function claimSeat(System $system, User $player, string $seat_type
...
+        } else {
+            $message = "No village seat found!";
+        }
+        // if active challenges to your seat, auto win for the challenger
+        if (isset($challenger)) {
+            // verify challenger meets requirements
+            self::checkSeatRequirements($system, $challenger, $player_seat_type);
+            // claim seat for challenger
+            self::claimSeat($system, $challenger, $player_seat_type);
+        }
```
levimeahan, line 394: Hmm, what if there are multiple challengers?

levimeahan: Ah I see the query above pulls the first challenge request 👍

classes/village/VillageManager.php
```diff
@@ -347,25 +347,74 @@ public static function claimSeat(System $system, User $player, string $seat_type
...
      */
     public static function resign(System $system, User $player): string
     {
+        $message = '';
+        $time = time();
+        $player_seat = $player->village_seat;
+        $player_seat_id = $player_seat->seat_id;
+        $player_seat_type = $player_seat->seat_type;
+        // get challenger if set
+        $challenger = null;
```
levimeahan, line 360:
```suggestion
        $first_challenger = null;
```
to make the intended behavior here more clear

levimeahan (approved):

classes/village/VillageManager.php
aumu1031: > to make the intended behavior here more clear

Done!

PR #600, opened 2024-01-28
Fix Duplicate Votes

aumu1031:


db/migrations/20240128161550_duplicate_vote_migration.php
```diff
@@ -0,0 +1,29 @@
...
+final class DuplicateVoteMigration extends AbstractMigration
+{
+    /**
+     * Migrate Up.
+     */
+    public function up(): void
+    {
+        $this->execute("
+            -- Alter table vote_logs
+            ALTER TABLE `vote_logs` ADD CONSTRAINT `unique_user_proposal_combination` UNIQUE (user_id, proposal_id);
```
levimeahan, line 15: Is it possible this migration could fail due to having non-unique rows in the table?

aumu1031: I manually cleared those rows previously where the bug was reported to me but I'll do a sweep before merging.

levimeahan (approved):

PR #616, opened 2024-02-04
Modal Component

aumu1031:
Step 1: import
image
Step 2: use hook
image
Step 3: apply modal context via ModalProvider
image
Step 4: open modal
image


levimeahan (approved):
Sweet!

PR #623, opened 2024-02-14
Village Upgrade Classes

aumu1031:
Dto classes and VillageUpgradeManager

### DIFF
--- a/classes/village/VillageUpgradeDto.php
+++ b/classes/village/VillageUpgradeDto.php
@@ -1,0 +1,25 @@
+<?php
+
+class VillageUpgradeDto {
+    public function __construct(
+        public ?int $id = null,
+        public string $key,
+        public int $village_id,
+        public string $status = VillageUpgradeConfig::UPGRADE_STATUS_LOCKED,
+        public ?int $research_progress = null,
+        public ?int $research_progress_required = null,
+        public ?int $research_progress_last_updated = null,
+        // display-only properties
+        public string $name = "",
+        public string $description = "",
+        public ?int $materials_research_cost = null,
+        public ?int $food_research_cost = null,
+        public ?int $wealth_research_cost = null,
+        public ?int $research_time = null,
+        public ?string $research_time_remaining = null,
+        public ?int $materials_upkeep = null,
+        public ?int $food_upkeep = null,
+        public ?int $wealth_upkeep = null,
+        public ?int $requirements_met = null,
+    ) {}
+}
--- a/classes/village/VillageUpgradeManager.php
+++ b/classes/village/VillageUpgradeManager.php
@@ -1,0 +1,368 @@
+<?php
+
+require_once __DIR__ . "/VillageBuildingDto.php";
+require_once __DIR__ . "/VillageUpgradeDto.php";
+require_once __DIR__ . "/VillageUpgradeSetDto.php";
+require_once __DIR__ . "/VillageBuildingConfig.php";
+require_once __DIR__ . "/VillageUpgradeConfig.php";
+
+class VillageUpgradeManager {
+    const UPGRADE_TOGGLE_COOLDOWN_DAYS = 3;
+
+    /**
+     * @param System  $system
+     * @param int $village_id
+     * @return VillageBuildingDto[]
+     */
+    public static function getBuildingsForVillage(System $system, int $village_id): array {
+        $buildings = [];
+        $query = $system->db->query("
+            SELECT * FROM `village_buildings` WHERE `village_id` = {$village_id}
+        ");
+        $building_data = $system->db->fetch_all($query);
+        foreach ($building_data as $building) {
+            $buildings[$building['key']] = new VillageBuildingDto(
+                id: $building['id'],
+                key: $building['key'],
+                village_id: $building['village_id'],
+                tier: $building['tier'],
+                health: $building['health'],
+                max_health: VillageBuildingConfig::BUILDING_MAX_HEALTH[$building['key']][$building['tier']],
+                defense: $building['defense'],
+                status: $building['status'],
+                construction_progress: $building['construction_progress'],
+                construction_progress_required: $building['construction_progress_required'],
+                construction_progress_last_updated: $building['construction_progress_last_updated'],
+            );
+        }
+        return $buildings;
+    }
+
+    /**
+     * @param System  $system
+     * @param int $village_id
+     * @return VillageUpgradeDto[]
+     */
+    public static function getUpgradesForVillage(System $system, int $village_id): array {
+        $upgrades = [];
+        $query = $system->db->query("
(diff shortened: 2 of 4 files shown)


classes/village/VillageUpgradeSetDto.php
aumu1031, line 1: Note this class exists solely to define an intuitive structure for the front end to consume

classes/village/VillageUpgradeManager.php
```diff
@@ -0,0 +1,368 @@
...
+                        food_research_cost: VillageUpgradeConfig::UPGRADE_RESEARCH_COST[$upgrade_key][WarManager::RESOURCE_FOOD],
+                        wealth_research_cost: VillageUpgradeConfig::UPGRADE_RESEARCH_COST[$upgrade_key][WarManager::RESOURCE_WEALTH],
+                        research_time: VillageUpgradeConfig::UPGRADE_RESEARCH_TIME[$upgrade_key],
+                        research_time_remaining: isset($village->upgrades[$upgrade_key]->research_progress_last_updated) ? System::TimeRemaining($village->upgrades[$upgrade_key]->research_progress_required - $village->upgrades[$upgrade_key]->research_progress, format: "long", include_seconds: false, include_minutes: false) : null,
+                        food_upkeep: VillageUpgradeConfig::UPGRADE_UPKEEP[$upgrade_key][WarManager::RESOURCE_FOOD],
+                        materials_upkeep: VillageUpgradeConfig::UPGRADE_UPKEEP[$upgrade_key][WarManager::RESOURCE_MATERIALS],
+                        wealth_upkeep: VillageUpgradeConfig::UPGRADE_UPKEEP[$upgrade_key][WarManager::RESOURCE_WEALTH],
+                        requirements_met: VillageUpgradeManager::checkResearchRequirementsMet($village, $upgrade_key),
```
levimeahan, line 169: Had an idea here:
- Split `VillageUpgradeDto` into two classes:
  - `VillageUpgrade` represents the config + the current status for this village. This is the main class we work with in application code. The idea is it's never missing data, if you have an instance of `VillageUpgrade`, you have everything you need
  - `VillageUpgradeDb` / `VillageUpgradeData` represents just the status for this village loaded from the db in `getUpgradesForVillage`
- Move this initialize logic to a helper in `VillageUpgrade`

```php

public static function getBuildingUpgradesForDisplay(System $system, Village $village): array {
  // ..
  foreach (VillageBuildingConfig::UPGRADE_SET_UPGRADES[$upgrade_set_key] as $upgrade_key) {
    $upgrade = VillageUpgrade::fromUpgradeData(
      key: $upgrade_key, 
      village_id: $village->id, 
      upgrade_data: $village->upgrade_data[$upgrade_key]
    );
  }
  // ..
}

// ...

class VillageUpgrade {
  // ...
  public static function init(string $upgrade_key, int $village_id, VillageUpgradeData $upgrade_data): VillageUpgrade {
      return new VillageUpgradeDto(
      id: $upgrade_data->id ?? null,
      key: $upgrade_key,
      village_id: $village_id,
      status: $upgrade_data->status ?? VillageUpgradeConfig::UPGRADE_STATUS_LOCKED,
      // ..
      name: VillageUpgradeConfig::UPGRADE_NAMES[$upgrade_key],
      // ..
  }
}
```

The main benefit I'm thinking this achieves is "if you have an instance of `VillageUpgrade`, you know that the values are there" rather than sometimes having an instance that doesn't have the config values set (from `getUpgradesForVillage`).

Side benefit is isolating some of the complexity out of this method that is more related to how config maps to a `VillageUpgrade` instance, so this class is slightly easier to read through

classes/village/VillageUpgradeDto.php
```diff
@@ -0,0 +1,25 @@
...
+        public string $name = "",
+        public string $description = "",
+        public ?int $materials_research_cost = null,
+        public ?int $food_research_cost = null,
+        public ?int $wealth_research_cost = null,
+        public ?int $research_time = null,
+        public ?string $research_time_remaining = null,
+        public ?int $materials_upkeep = null,
+        public ?int $food_upkeep = null,
+        public ?int $wealth_upkeep = null,
```
levimeahan, line 22: I have comments below about this but the "all these values can potentially be null/empty" is something I think we can potentially use two versions of this class for

- One completely without these fields, for the cases where they are not present
- One with these fields marked as not nullable, for the cases where they are present

The `research_time_remaining` is a bit different where it's totally valid for that variable to independently be null or not null depending on whether or not there's research happening going on. But some of these others I think are more grouped around "was config loaded or not" and that's a pattern I think having a different class for can help with. See below for more detailed suggestion

classes/village/VillageUpgradeManager.php
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    /**
+     * @param System  $system
+     * @param VillageUpgradeDto[] $upgrades
+     * @return array
+     */
+    public static function initializeEffectsForVillage(System $system, array $upgrades): array {
+        // initialize array with defaults
+        $effects = [
```
levimeahan, line 73: Should this array of "what all effects are there" be defined in `VillageUpgradeConfig` instead and we can clone it in here and initialize all values to 0? Might reduce instances of adding a new effect and forgetting to update this list

classes/village/VillageUpgradeManager.php
```diff
@@ -0,0 +1,368 @@
...
+            if ($village->resources[WarManager::RESOURCE_MATERIALS] < $materials_cost || $village->resources[WarManager::RESOURCE_FOOD] < $food_cost || $village->resources[WarManager::RESOURCE_WEALTH] < $wealth_cost) {
+                return "Not enough resources!";
+            }
+            // update village resources
+            $village->subtractResource(WarManager::RESOURCE_MATERIALS, $materials_cost);
+            $village->subtractResource(WarManager::RESOURCE_FOOD, $food_cost);
+            $village->subtractResource(WarManager::RESOURCE_WEALTH, $wealth_cost);
+            $village->updateResources();
+            // log expenditure in resource_logs table
+            $system->db->query("INSERT INTO `resource_logs`(`village_id`, `resource_id`, `type`, `quantity`, `time`) VALUES ({$village->village_id}, " . WarManager::RESOURCE_MATERIALS . ", " . VillageManager::RESOURCE_LOG_CONSTRUCTION_COST . ", {$materials_cost}, " . time() . ")");
```
levimeahan, line 267: Could/should we move this into `$village->subtractResource`? Is there a valid case for subtracting a resource without logging it?

levimeahan (approved):

classes/village/VillageUpgradeManager.php
aumu1031: For the cron job we want the ability to get a query string, I wonder if we make subtract/addResource return a query string, or alternatively take an optional parameter that executes it automatically (updateResources works this way).

PR #626, opened 2024-02-15
War Table Updates

aumu1031:
image

### DIFF
--- a/classes/village/VillageApiPresenter.php
+++ b/classes/village/VillageApiPresenter.php
@@ -120,7 +120,7 @@ class VillageApiPresenter {
             VillageManager::getChallengeData($system, $player)
         );
     }
-    public static function warLogDataResponse(System $system, User $player, int $page_number = 1): array {
+    public static function playerWarLogDataResponse(System $system, User $player, int $page_number = 1): array {
         $global_leaderboard_war_logs = array_map(
             function (WarLogDto $war_log) {
                 return [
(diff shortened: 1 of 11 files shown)


classes/village/VillageApiPresenter.php
```diff
@@ -161,6 +161,84 @@ function (WarLogDto $war_log) {
...
+                        "villages_captured" => $warRecordDto->attacker_war_log->villages_captured,
+                        "pvp_wins" => $warRecordDto->attacker_war_log->pvp_wins,
+                        "points_gained" => $warRecordDto->attacker_war_log->points_gained,
+                        "stability_gained" => $warRecordDto->attacker_war_log->stability_gained,
+                        "stability_reduced" => $warRecordDto->attacker_war_log->stability_reduced,
+                        "war_score" => $warRecordDto->attacker_war_log->war_score,
+                        "objective_score" => $warRecordDto->attacker_war_log->objective_score,
+                        "resource_score" => $warRecordDto->attacker_war_log->resource_score,
+                        "battle_score" => $warRecordDto->attacker_war_log->battle_score,
+                    ],
```
levimeahan, line 205: This is a great candidate for extracting to its own method, could then be reused for the defender war log below as well
```suggestion
                    "attacker_war_log" => VillageApiPresenter::warLogResponse($warRecordDto->attacker_war_log),
```

classes/village/VillageApiPresenter.php
```diff
@@ -161,6 +161,84 @@ function (WarLogDto $war_log) {
             "player_war_log" => $player_war_log
         ];
     }
+    public static function warRecordDataResponse(System $system, User $player, int $page_number = 1): array
+    {
+        $war_records = array_map(
+            function (WarRecordDto $warRecordDto) {
+                return [
```
levimeahan, line 168: Not a big deal but if we're immediately returning the value of `$war_records`, we could skip a step and return it directly
```suggestion
        return [
            "war_records" => array_map(
                function (WarRecordDto $warRecordDto) {
                    return [
```

classes/village/VillageApiPresenter.php
```diff
@@ -161,6 +161,84 @@ function (WarLogDto $war_log) {
             "player_war_log" => $player_war_log
         ];
     }
+    public static function warRecordDataResponse(System $system, User $player, int $page_number = 1): array
```
levimeahan, line 164: - This returns multiple war records so we could make it plural
- everything in these responses is kinda "data" so maybe we can drop that word from the naming
```suggestion
    public static function warRecordsResponse(System $system, User $player, int $page_number = 1): array
```

ui_components/src/village/Village.js
```diff
@@ -212,10 +213,13 @@ function Village({
             }
             {villageTab == "warTable" &&
                 <WarTable
-                warLogData={warLogData}
+                playerWarLogData={playerWarLogData}
+                warRecordData={warRecordData}
```
levimeahan, line 217: Maybe same comment as the presenter here - Is "data" adding something meaningful to these names?

ui_components/src/village/WarTable.js
```diff
@@ -121,6 +127,183 @@ export function WarTable({
...
+                        <div className="village_warlog_details_value">{log.patrols_defeated}</div>
+                    </div>
+                    <div className="village_warlog_details_item">
+                        <div className="village_warlog_details_label">RESOURCES</div>
+                        <div className="village_warlog_details_value">{log.resources_stolen}</div>
+                    </div>
+                </div>
+            </div>
+        );
+    }
```
levimeahan, line 305: Can you move these components down below the function body of `WarTable`? Will help with reading through what WarTable itself is doing

ui_components/src/village/WarTable.js
```diff
@@ -208,6 +429,66 @@ export function WarTable({
...
+                                    <feFuncA type="linear" slope="1" />
+                                </feComponentTransfer>
+                                <feComposite in="floodColor" in2="opacityAdjustedBlur" operator="in" result="coloredBlur" />
+                                <feMerge>
+                                    <feMergeNode in="coloredBlur" />
+                                    <feMergeNode in="SourceGraphic" />
+                                </feMerge>
+                            </filter>
+                        </defs>
+                    </svg>
```
levimeahan, line 449: This looks like a good candidate to be its own component

levimeahan (approved):

ui_components/src/village/Village.js
aumu1031: Yeah representing that these values are for reference and not manipulation unless as part of a server response.

ui_components/src/village/WarTable.js
aumu1031: Yep makes sense, we only really care about the color, slope, and deviation for the sake of how the output looks so those can be a parameter for a `GlowUtil` or similarly named utility.

ui_components/src/village/WarTable.js
aumu1031: Done.

classes/village/VillageApiPresenter.php
aumu1031: Keeping it 1:1 with the name used in the template file and js, explanation for specifying "data" in the other comment.

classes/village/VillageApiPresenter.php
aumu1031: Yup makes sense.

PR #629, opened 2024-02-17
v0.12c Back End

aumu1031:


classes/village/VillageApiPresenter.php
```diff
@@ -255,4 +200,104 @@ function ($row) {
...
+                                            "food_research_cost" => $villageUpgradeDto->food_research_cost,
+                                            "wealth_research_cost" => $villageUpgradeDto->wealth_research_cost,
+                                            "research_time" => $villageUpgradeDto->research_time,
+                                            "research_time_remaining" => $villageUpgradeDto->research_time_remaining,
+                                            "food_upkeep" => $villageUpgradeDto->food_upkeep,
+                                            "materials_upkeep" => $villageUpgradeDto->materials_upkeep,
+                                            "wealth_upkeep" => $villageUpgradeDto->wealth_upkeep,
+                                            "requirements_met" => $villageUpgradeDto->requirements_met,
+                                        ];
+                                    },
```
levimeahan, line 260: This looks like a good candidate to be its own presenter and keep this function from being too large

classes/village/VillageBuildingConfigData.php
```diff
@@ -0,0 +1,50 @@
...
+        return $this->phrase;
+    }
+    public function getBackgroundImage(): string {
+        return $this->background_image;
+    }
+    public function getMaxHealth($tier): int {
+        return $this->max_healths[$tier] ?? 0;
+    }
+    public function getConstructionCostMaterials($tier): int {
+        return $this->construction_costs[$tier][WarManager::RESOURCE_MATERIALS] ?? 0;
```
levimeahan, line 36: This and similar functions should probably throw a RuntimeException if its an invalid tier so we don't inadvertently allow free construction if we messed up the tier we passed into this check

classes/village/VillageBuildingConfig.php
```diff
@@ -35,263 +59,275 @@ class VillageBuildingConfig {
...
+    /* construction cost indexed by building key, then tier, then resource type */
     const BUILDING_CONSTRUCTION_COST = [
         self::BUILDING_VILLAGE_HQ => [
             1 => [
-                WarManager::RESOURCE_MATERIALS => 10800,
-                WarManager::RESOURCE_FOOD => 4320,
-                WarManager:: RESOURCE_WEALTH => 6480,
+                WarManager::RESOURCE_MATERIALS => self::BASE_T1_CONSTRUCTION_COST_TOTAL * 0.5 * 1.5,
+                WarManager::RESOURCE_FOOD => self::BASE_T1_CONSTRUCTION_COST_TOTAL * 0.2 * 1.5,
+                WarManager::RESOURCE_WEALTH => self::BASE_T1_CONSTRUCTION_COST_TOTAL * 0.3 * 1.5,
```
levimeahan, line 156: What is the `* 1.5` modifier representing here?

classes/village/VillageBuildingDto.php
```diff
@@ -13,16 +13,17 @@ public function __construct(
...
-        public string $description = "",
-        public string $phrase = "",
-        public ?int $materials_construction_cost = null,
-        public ?int $food_construction_cost = null,
-        public ?int $wealth_construction_cost = null,
-        public ?int $construction_time = null,
-        public ?string $construction_time_remaining = null,
-        public ?int $requirements_met = null,
+        public bool $construction_boosted = false,
+        public string $construction_time_remaining = '',
```
levimeahan, line 17: What's the expected value of this field? Should we use int or a date type instead?

classes/village/VillageManager.php
```diff
@@ -518,49 +532,66 @@ public static function getVillageSeats(System $system, int $village_id): array {
...
-            $result = $system->db->fetch($result);
-            $resource_history[$resource_id]['lost'] = (int) $result['lost_loot_count'];
-            // get upkeep - WIP
-            $result = $system->db->query("SELECT SUM(`quantity`) as 'spent' FROM `resource_logs` WHERE `resource_id` = {$resource_id} AND `village_id` = {$village_id} AND `type` = " . self::RESOURCE_LOG_EXPENSE . " AND `time` > {$time}");
-            $result = $system->db->fetch($result);
-            $resource_history[$resource_id]['spent'] = (int) $result['spent'];
+        $query = "
+            SELECT
+                `resource_id`,
+                SUM(CASE WHEN `type` = " . self::RESOURCE_LOG_PRODUCTION . " THEN `quantity` ELSE 0 END) AS 'produced',
```
levimeahan, line 553: Just checking, do we have an index on `type` in the resource_logs table? How often will this function be called?

classes/village/VillageManager.php
```diff
@@ -1589,15 +1625,6 @@ public static function enactProposal(System $system, User $player, int $proposal
...
                 self::createProposalNotification($system, $proposal['village_id'], NotificationManager::NOTIFICATION_PROPOSAL_PASSED, $proposal['name']);
-                // if village occupied by enemy, return to owner's control
-                $system->db->query("UPDATE `region_locations`
-                    INNER JOIN `regions` on `regions`.`region_id` = `region_locations`.`region_id`
-                    SET `region_locations`.`occupying_village_id` = NULL
-                    WHERE `regions`.`village` = {$proposal['target_village_id']} AND `region_locations`.`occupying_village_id` = {$proposal['village_id']}");
-                $system->db->query("UPDATE `region_locations`
-                    INNER JOIN `regions` on `regions`.`region_id` = `region_locations`.`region_id`
-                    SET `region_locations`.`occupying_village_id` = NULL
-                    WHERE `regions`.`village` = {$proposal['village_id']} AND `region_locations`.`occupying_village_id` = {$proposal['target_village_id']}");
```
levimeahan, line 1600: What's the motivation for removing this behavior?

classes/village/VillageManager.php
```diff
@@ -1708,6 +1726,72 @@ public static function enactProposal(System $system, User $player, int $proposal
...
+            case self::PROPOSAL_TYPE_BEGIN_CONSTRUCTION:
+                try {
+                    $message = VillageUpgradeManager::beginConstruction($system, $player->village, $proposal['building_key']);
+                    // update proposal
+                    $system->db->query("UPDATE `proposals` SET `end_time` = {$time}, `result` = 'passed' WHERE `proposal_id` = {$proposal['proposal_id']}");
+                    // create notification
+                    self::createProposalNotification($system, $proposal['village_id'], NotificationManager::NOTIFICATION_PROPOSAL_PASSED, $proposal['name']);
+                } catch (RuntimeException $e) {
+                    return $e->getMessage();
+                }
```
levimeahan, line 1738: This logic is duplicated for every proposal in here and this function is getting super long - Could you move the try/catch above the switch statement, and also the lines for updating the proposal in DB and creating notification? 

Could also make the proposal implementation functions like `VillageUpgradeManager::beginConstruction` return an instance of `ActionResult` and then you can handle success/failure in a standard way

classes/village/VillageManager.php
```diff
@@ -2255,10 +2585,10 @@ private static function setNewRelations(System $system, int $initiator_village_i
...
                 case VillageRelation::RELATION_ALLIANCE:
-                    $relation_name = VillageManager::VILLAGE_NAMES[$initiator_village_id] . "-" . VillageManager::VILLAGE_NAMES[$recipient_village_id] . " Alliance";
+                    $relation_name = VillageManager::VILLAGE_NAMES[$initiator_village_id] . " - " . VillageManager::VILLAGE_NAMES[$recipient_village_id] . " Alliance";
                     break;
                 case VillageRelation::RELATION_WAR:
-                    $relation_name = VillageManager::VILLAGE_NAMES[$initiator_village_id] . "-" . VillageManager::VILLAGE_NAMES[$recipient_village_id] . " War";
+                    $relation_name = VillageManager::VILLAGE_NAMES[$initiator_village_id] . " - " . VillageManager::VILLAGE_NAMES[$recipient_village_id] . " War";
                     break;
                 default:
                     break;
```
levimeahan, line 2594: At 2594 lines I think it's time to start thinking about if we can split this into two classes

classes/village/VillageUpgrade.php
```diff
@@ -0,0 +1,51 @@
...
+        return $this->config_data->getName();
+    }
+    public function getTier(): int {
+        return $this->config_data->getTier();
+    }
+    public function getDescription(): string {
+        return $this->config_data->getDescription();
+    }
+    public function getResearchTime(): int {
+        return $this->config_data->getResearchTime();
```
levimeahan, line 25: Where are you planning to account for research boosts?

classes/village/VillageUpgradeManager.php
```diff
@@ -134,55 +147,82 @@ public static function initializeEffectsForVillage(System $system, array $upgrad
...
+                        research_time_remaining: $upgrade ? System::TimeRemaining($upgrade->research_progress_required - $upgrade->research_progress, format: "long", include_seconds: false, include_minutes: false) : '',
+                        name: $upgrade_config_data->getName(),
+                        tier: $upgrade_config_data->getTier(),
+                        description: $upgrade_config_data->getDescription(),
+                        materials_research_cost: ($upgrade && $upgrade->research_progress !== null) ? 0 : $upgrade_config_data->getResearchCostMaterials(),
+                        food_research_cost: ($upgrade && $upgrade->research_progress !== null) ? 0 : $upgrade_config_data->getResearchCostFood(),
+                        wealth_research_cost: ($upgrade && $upgrade->research_progress !== null) ? 0 : $upgrade_config_data->getResearchCostWealth(),
+                        research_time: ($upgrade && $upgrade->research_progress !== null) ? round($upgrade_config_data->getResearchTime() * (($upgrade->research_progress_required - $upgrade->research_progress) / $upgrade->research_progress_required)) : $upgrade_config_data->getResearchTime(),
```
levimeahan, line 175: this formatting makes it easier to read long ternaries
```suggestion
                        research_time: ($upgrade && $upgrade->research_progress !== null) 
                            ? round($upgrade_config_data->getResearchTime() * (($upgrade->research_progress_required - $upgrade->research_progress) / $upgrade->research_progress_required)) 
                            : $upgrade_config_data->getResearchTime(),
```

classes/village/VillageUpgradeManager.php
```diff
@@ -234,29 +277,31 @@ public static function checkConstructionRequirementsMet(Village $village, string
...
+        self::initialize();
         // get construction costs and time
         $materials_cost = VillageBuildingConfig::BUILDING_CONSTRUCTION_COST[$building_key][$village->buildings[$building_key]->tier + 1][WarManager::RESOURCE_MATERIALS];
         $food_cost = VillageBuildingConfig::BUILDING_CONSTRUCTION_COST[$building_key][$village->buildings[$building_key]->tier + 1][WarManager::RESOURCE_FOOD];
         $wealth_cost = VillageBuildingConfig::BUILDING_CONSTRUCTION_COST[$building_key][$village->buildings[$building_key]->tier + 1][WarManager::RESOURCE_WEALTH];
         $progress_required = VillageBuildingConfig::BUILDING_CONSTRUCTION_TIME[$building_key][$village->buildings[$building_key]->tier + 1] * 86400;
         // check if requirements met
         if (!VillageUpgradeManager::checkConstructionRequirementsMet($village, $building_key, $village->buildings[$building_key]->tier + 1)) {
-            return "Construction requirements not met!";
+            throw new RuntimeException("Construction requirements not met!");
```
levimeahan, line 292: Per my other comment, optionally you can return an instance of https://github.com/elementum-games/shinobi-chronicles/blob/main/classes/ActionResult.php as a structured way to handle this that doesn't have to be try/caught. For an examples see `User#useJutsu()` 

https://github.com/elementum-games/shinobi-chronicles/blob/8bf516d26c8538b0f52eb5dee737fa987e5dadc6/classes/User.php#L1634

https://github.com/elementum-games/shinobi-chronicles/blob/8bf516d26c8538b0f52eb5dee737fa987e5dadc6/classes/SpecialMission.php#L576-L579

Up to you but personally I like the structure of it and avoiding the `try/catch` everywhere

levimeahan (approved):

classes/village/VillageBuildingDto.php
aumu1031: `$construction_time_remaining` is a formatted display string.

classes/village/VillageBuildingConfig.php
aumu1031: Making some buildings have a higher or lower total construction time/cost. VillageHQ for example, 50% more expensive while Ramen Stand is 50% less expensive.

classes/village/VillageManager.php
aumu1031: We do not yet but that's part of optimizing this query. Code-wise changing this to hitting the logs table once rather than multiple times per resource, and then adding an index on type.

classes/village/VillageManager.php
aumu1031: I don't think we strictly want to remove this functionality just yet, but its part of decoupling town control from region control. So even if you trade a region you control to someone else it doesn't steal town control from someone else.

classes/village/VillageManager.php
aumu1031: I'm planning on taking a pass at proposal logic for this reason. There's a lot of code reuse and VillageManager is getting bloated, so I want to make a ProposalManager class to handle all of this logic and cut down on repeated actions.

For ActionResult vs Try/Catch with a String return, which do you think is better? ActionResult would have a string message and bool success. Whichever we choose I would like to apply to the other functions here.

classes/village/VillageUpgrade.php
aumu1031: Research time is just the config value, so research boosts will be factored in two places:
- VillageUpgradeDto which passes a display string for the remaining time, we'll factor the village's research speed into that calculation and for the progress bar display on the front end.
- HourlyCron factors in research speed as a multiplier to the progress gained.

classes/village/VillageUpgradeManager.php
aumu1031: I think I'll even make this a separate calculation before the class constructor.

PR #632, opened 2024-02-20
Fix View Battles Resid

aumu1031:


pages/viewBattles.php
```diff
@@ -366,7 +366,10 @@ function viewBattles() {
                 if ($system->db->last_num_rows > 0) {
                     while ($turn = $system->db->fetch($logs_result)) {
                         $battle_text = $system->html_parse(stripslashes($turn['content']));
-                        $battle_text = str_replace(array('[br]', '[hr]'), array('<br />', '<hr />'), $battle_text);
+                        $battle_text = str_replace(
```
levimeahan, line 369: Hmm this seems like a good time to add this to an array somewhere in case we want to make more changes, maybe on `BattleManager`?

levimeahan (approved):

PR #636, opened 2024-02-20
Display elements on profile

Algharna:
add elements to display on profile
moved around a few rows
added number_format
spouse no longer displays if user has no spouse


aumu1031:
I recommend including a screenshot of the UI changes you want to make so its possible to see what the changes look like without needing to pull this first. I assume you didn't mean to include the docker/gitignore files, so if possible try to pick which files you want to include when you push to your fork.

PR #640, opened 2024-02-22
v0.12 Upgrades Back End

aumu1031:


classes/village/VillageManager.php
```diff
@@ -1708,6 +1726,72 @@ public static function enactProposal(System $system, User $player, int $proposal
                 self::createProposalNotification($system, $proposal['village_id'], NotificationManager::NOTIFICATION_PROPOSAL_PASSED, $proposal['name']);
                 $message = "Accepted trade offer from " . VillageManager::VILLAGE_NAMES[$proposal['target_village_id']] . "!";
                 break;
+            case self::PROPOSAL_TYPE_BEGIN_CONSTRUCTION:
```
aumu1031, line 1729: Will shift this entire code block and sub-functions to `ProposalManager` and use `ActionResult` instead of try/catch

classes/village/VillageManager.php
```diff
@@ -2177,11 +2504,12 @@ private static function checkTradeValid(System $system, int $village1_id, int $v
         return true;
     }
 
-    private static function handleTradeCompletion(System $system, int $village1_id, int $village2_id, array $offered_resources, array $offered_regions, array $requested_resources, array $requested_regions)
-    {
+    private static function handleTradeCompletion(System $system, int $village1_id, int $village2_id, array $offered_resources, array $offered_regions, array $requested_resources, array $requested_regions) {
         // update regions
         foreach ($offered_regions as $region) {
             $system->db->query("UPDATE `regions` SET `village` = {$village2_id} WHERE `region_id` = {$region}");
+            // update castles and towns in region if available to trade
```
aumu1031, line 2511: Only trade towns in a region if they're currently controlled.

classes/Village.php
aumu1031, line 1: Thinking we can turn `activeUpgradeEffects[]` into a sub-class like `$village->policy` instead of having to get upgrade effects by key. So that way all of the upgrade/building configuration will be accessed via classes rather than needing to interact directly with anything in the configuration file.

classes/Village.php
```diff
@@ -38,6 +58,34 @@ public function __construct($system, $village = '', array $village_row = []) {
             $this->coords = VillageManager::getLocation($this->system, $this->village_id);
             $this->relations = VillageManager::getRelationsForVillage($this->system, $this->village_id);
             $this->policy = new VillagePolicy($this->policy_id);
+            $this->buildings = VillageUpgradeManager::getBuildingsForVillage($this->system, $this->village_id);
+            $this->upgrades = VillageUpgradeManager::getUpgradesForVillage($this->system, $this->village_id);
+            $this->active_upgrade_effects = VillageUpgradeManager::initializeEffectsForVillage($this->system, $this->upgrades);
+            $this->construction_speed += ($this->active_upgrade_effects[VillageUpgradeConfig::UPGRADE_EFFECT_CONSTRUCTION_SPEED] / 100);
+            $this->research_speed += ($this->active_upgrade_effects[VillageUpgradeConfig::UPGRADE_EFFECT_RESEARCH_SPEED] / 100);
+            $this->resource_capacity += $this->active_upgrade_effects[VillageUpgradeConfig::UPGRADE_EFFECT_RESOURCE_CAPACITY];
+            if ($this->policy->construction_speed > 0) {
```
aumu1031, line 67: This policy benefits to research/construction are based on how many villages have more total research/construction.

classes/Village.php
```diff
@@ -47,6 +95,11 @@ public function __construct($system, $village = '', array $village_row = []) {
             $this->coords = VillageManager::getLocation($this->system, $this->village_id);
             $this->relations = VillageManager::getRelationsForVillage($this->system, $this->village_id);
             $this->policy = new VillagePolicy($this->policy_id);
+            $this->buildings = VillageUpgradeManager::getBuildingsForVillage($this->system, $this->village_id);
```
aumu1031, line 98: Need to combine this constructor logic.

classes/village/VillageManager.php
```diff
@@ -518,49 +532,66 @@ public static function getVillageSeats(System $system, int $village_id): array {
...
-            $resource_history[$resource_id]['claimed'] = (int)$result['claimed_loot_count'];
-            // get lost (lol)
-            $result = $system->db->query("SELECT COUNT(*) as 'lost_loot_count' FROM `loot` WHERE `resource_id` = {$resource_id} AND `target_village_id` = {$village_id} AND (`claimed_village_id` != {$village_id} OR `claimed_village_id` IS NULL) AND (`claimed_time` > {$time} OR `claimed_time` IS NULL)");
-            $result = $system->db->fetch($result);
-            $resource_history[$resource_id]['lost'] = (int) $result['lost_loot_count'];
-            // get upkeep - WIP
-            $result = $system->db->query("SELECT SUM(`quantity`) as 'spent' FROM `resource_logs` WHERE `resource_id` = {$resource_id} AND `village_id` = {$village_id} AND `type` = " . self::RESOURCE_LOG_EXPENSE . " AND `time` > {$time}");
-            $result = $system->db->fetch($result);
-            $resource_history[$resource_id]['spent'] = (int) $result['spent'];
+        $query = "
```
aumu1031, line 550: This changes to a single query rather than multiple queries for multiple resources. Used only for the village HQ display.

classes/village/VillageUpgradeManager.php
```diff
@@ -681,4 +715,26 @@ public static function boostResearch(System $system, Village $village, string $u
         $system->db->query("UPDATE `village_upgrades` SET `research_boosted` = 1, `research_progress_last_updated` = " . time() . ", `research_progress_required` = " . $village->upgrades[$upgrade_key]->research_progress_required . " WHERE `id` = {$village->upgrades[$upgrade_key]->id}");
         return "Research boosted for " . VillageUpgradeConfig::UPGRADE_NAMES[$upgrade_key] . "!";
     }
+
+    public static function calcResearchScore(System $system, Village $village): int {
```
aumu1031, line 719: These are a count of total days completed construction/research time, currently just for the catch-up bonus with Ashes policy but also a useful way for us to glance at the DB and see how much progress villages are making.

classes/Village.php
levimeahan: sounds good!

classes/Village.php
```diff
@@ -38,6 +58,34 @@ public function __construct($system, $village = '', array $village_row = []) {
...
+                $result = $system->db->query("SELECT `village_id`, `construction_score` FROM `villages`");
+                $village_scores = $system->db->fetch_all($result);
+                $num_villages = 0;
+                foreach ($village_scores as $score) {
+                    if ($score['village_id'] != $this->village_id && $score['construction_score'] > $this->construction_score) {
+                        $num_villages++;
+                    }
+                }
+                $this->construction_speed += ($num_villages * $this->policy->construction_speed) / 100;
+            }
```
levimeahan, line 77: Consider extracting this to a method and then doing something like `$this->construction_speed = $this->getConstructionSpeed()`

classes/Village.php
```diff
@@ -47,6 +95,33 @@ public function __construct($system, $village = '', array $village_row = []) {
             $this->coords = VillageManager::getLocation($this->system, $this->village_id);
             $this->relations = VillageManager::getRelationsForVillage($this->system, $this->village_id);
             $this->policy = new VillagePolicy($this->policy_id);
+            $this->buildings = VillageUpgradeManager::getBuildingsForVillage($this->system, $this->village_id);
+            $this->upgrades = VillageUpgradeManager::getUpgradesForVillage($this->system, $this->village_id);
+            $this->active_upgrade_effects = VillageUpgradeManager::initializeEffectsForVillage($this->system, $this->upgrades);
+            $this->construction_speed += $this->policy->construction_speed + $this->active_upgrade_effects[VillageUpgradeConfig::UPGRADE_EFFECT_CONSTRUCTION_SPEED];
+            $this->research_speed += $this->policy->research_speed + $this->active_upgrade_effects[VillageUpgradeConfig::UPGRADE_EFFECT_RESEARCH_SPEED];
+            if ($this->policy->construction_speed > 0) {
+                $result = $system->db->query("SELECT `village_id`, `construction_score` FROM `villages`");
```
levimeahan, line 104: one other note, we're making two separate queries to `villages` here, we could consider combining this into one query with the field that `construction_speed` and `research_speed` calculations use and pre-loading it into this class

classes/village/VillageUpgradeManager.php
```diff
@@ -156,6 +156,14 @@ public static function getBuildingUpgradesForDisplay(System $system, Village $vi
                 foreach (VillageBuildingConfig::UPGRADE_SET_UPGRADES[$upgrade_set_key] as $upgrade_key) {
                     $upgrade_config_data = VillageUpgradeManager::$UPGRADE_CONFIGS[$upgrade_key];
                     $upgrade = $village->upgrades[$upgrade_key] ?? null;
+                    if (isset($upgrade)) {
+                        if ($upgrade->research_progress != null) {
+                            $research_time_remaining = ($upgrade->research_progress_required - $upgrade->research_progress) / $village->research_speed;
+                            $research_time_remaining = System::TimeRemaining($research_time_remaining, format: "long", include_seconds: false, include_minutes: false);
```
levimeahan, line 162: Hmm, I wonder if there's a case for us to use the raw time value. Could consider storing the raw value in the DTO and then format it to a string where needed (can be a helper method on the `VillageUpgradeDTO` / `VillageUpgrade` class)

levimeahan (approved):
`ui_components/build` files are not supposed to be in the PR right? I'm okay merging as is and waiting on deploy to dev until the other PRs are merged

PR #641, opened 2024-02-22
v0.12 Travel

aumu1031:


ui_components/src/travel/Map.js
```diff
@@ -372,14 +373,41 @@ function RegionObjectives({ objectives, tileWidth, tileHeight, strategicView, pl
...
                         >
-                            <div className={'region_objective_tooltip' + (objective.is_occupied ? ' occupied' : '')} style={{ display: strategicView || (objective.x == player_x && objective.y == player_y) ? 'flex' : 'none' }}>
-                                <span className='region_objective_tooltip_name'>{objective.name}</span>
-                                <div className='region_objective_tooltip_tags'>
-                                    <span className='region_objective_tooltip_defense'>{objective.defense}</span>
-                                    <img className='region_objective_tooltip_village' src={getVillageIcon(objective.village_id)} />
-                                    <img className='region_objective_tooltip_resource' src={getResourceIcon(objective.resource_id)} />
-                                </div>
-                            </div>
+                            {(() => {
```
levimeahan, line 376: Ah this is curious, why is this function style necessary?

If it's to be able to use logic in rendering, considering pulling this part out into its own component

levimeahan (approved):

PR #644, opened 2024-02-22
v0.12 Upgrades Front End

aumu1031:


ui_components/src/village/VillageUpgrades.js
```diff
@@ -0,0 +1,754 @@
...
+            case 8:
+                return "VIII";
+            case 9:
+                return "IX";
+            case 10:
+                return "X";
+            default:
+                return "I";
+        }
+    }
```
levimeahan, line 68: This doesn't rely on any of the component's state or props, let's move it outside the component

ui_components/src/village/VillageUpgrades.js
```diff
@@ -0,0 +1,754 @@
...
+                ContentComponent: null,
+                onConfirm: null,
+            });
+        });
+    }
+    const buildingClickHandler = (building) => {
+        setSelectedUpgrade(null);
+        setSelectedBuilding(building);
+    };
+    const renderBuildingDetails = () => {
```
levimeahan, line 314: Could this function be its own component, with the data it needs passed in explicitly?

Same comment generally applies anytime you have a `renderX` function it may be a sign to pull that into a separate component (same file OK)

ui_components/src/village/VillageUpgrades.js
```diff
@@ -0,0 +1,754 @@
...
+                                    </div>
+                                }
+                            </div>
+                        </div>
+                    </div>
+                </div>
+            </div> 
+        );
+    }
+    const renderUpgradesContainer = () => {
```
levimeahan, line 513: Same comment here, this probably should be pulled out into its own component? And potentially some of the sub `render` functions as well

levimeahan (approved):

PR #645, opened 2024-02-24
number_format in various pages

Algharna:
Added number format to various places: 
profile page for pvp/ai wins
ramen shop heal amount/prices
pvp
image
image
image
image
image
battle log
image
view battle
image


aumu1031:
We had this convo a few times in chat, I think its aesthetically better without the number format in some places (like battle, bar #s) but there's definitely a subjective element to my preference.

levimeahan (approved):

PR #646, opened 2024-02-24
Cloudfalre img prep

HitoriWoB:
Allow T2 & T3 seal users to upload avatars directly to server.

This method requires the manual creation of `_user_imgs` in the base directory of the site.
This method will accept jpg, gif and png files that meet our file size requirements and store them as `{base_url}/_user_imgs/{user_name}.{file_type}`
Subsequent uploads will overwrite the existing image to save space.

Changed to only allow uploads once per day, to help avoid server stress.

### DIFF
--- a/pages/settings.php
+++ b/pages/settings.php
@@ -77,6 +77,11 @@ function userSettings() {
 			$file_type = strtolower(pathinfo($target_file, PATHINFO_EXTENSION));
 			$target_file_path = $target_dir . strtolower($player->user_name) . '.' . $file_type;
 			$user_link = $system->router->base_url . '_user_imgs/' . strtolower($player->user_name) . '.' . $file_type;
+
+			// Lack permission to upload file
+			if(!$player->forbidden_seal->direct_avatar_upload) {
+				throw new RuntimeException("You do not have permission to upload avatars!");
+			}
 	
 			// Upload directory is missing, this should be manually created
             if(!is_dir($target_dir)) {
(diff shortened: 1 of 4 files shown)


pages/settings.php
```diff
@@ -69,6 +69,69 @@ function userSettings() {
...
+
+			foreach($valid_file_types as $img_type) {
+				if(file_exists($target_dir . strtolower($player->user_name) . '.' . $img_type)) {
+					$current_file_path = $target_dir . strtolower($player->user_name) . '.' . $img_type;
+					break;
+				}
+			}
+			
+			// Limit uploads to 1/day
+			if($current_file_path && (time() - filemtime($current_file_path)) < 86400) {
```
levimeahan, line 91:
```suggestion
			if($current_file_path && (time() - filemtime($current_file_path)) < 60) {
```
Idk that this will create a ton of load, let's be more permissive here

pages/settings.php
```diff
@@ -69,6 +69,69 @@ function userSettings() {
...
+			if(!$player->forbidden_seal->direct_avatar_upload) {
+				throw new RuntimeException("You do not have permission to upload avatars!");
+			}
+	
+			// Upload directory is missing, this should be manually created
+            if(!is_dir($target_dir)) {
+                throw new RuntimeException("User upload directory is missing! Notify an administrator.");
+            }
+			// Only ally gif, png and jpg file types
+			if(!in_array($file_type, $valid_file_types)) {
```
levimeahan, line 105: we probably need to look at how to inspect files to ensure no malware even if they're named correctly

PR #651, opened 2024-03-17
Location check ramen

aumu1031:


classes/ramen_shop/RamenShopManager.php
```diff
@@ -475,6 +475,7 @@ public static function getMysteryRamen(User $player): MysteryRamenDto {
      * @return ActionResult
      */
     public static function purchaseBasicRamen(System $system, User $player, string $ramen_key): ActionResult {
+	if (self::checkLocationValid($system, $player)) {return ActionResult::failed("Invalid location!");}
```
levimeahan, line 478: Can leave off the braces if you're one-lining the if statement 👍
```suggestion
	if (self::checkLocationValid($system, $player)) return ActionResult::failed("Invalid location!");
```

levimeahan (approved):

PR #655, opened 2024-03-25
Initial duration display

aumu1031:


classes/village/VillageUpgradeManager.php
```diff
@@ -161,7 +161,8 @@ public static function getBuildingUpgradesForDisplay(System $system, Village $vi
                             $research_time_remaining = ($upgrade->research_progress_required - $upgrade->research_progress) / $village->research_speed;
                             $research_time_remaining = System::timeFormat($research_time_remaining, format: "long", include_seconds: false, include_minutes: false);
                         } else {
-                            $research_time_remaining = '';
+                            $research_time_remaining = $upgrade->research_progress_required / $village->research_speed;
+                            $research_time_remaining = System::timeFormat($research_time_remaining, format: "long", include_seconds: false, include_minutes: false);
```
levimeahan, line 165:
```suggestion
                            $research_time_remaining = $upgrade->research_progress_required / $village->research_speed;
```
FWIW I think it's more flexible if we store the number value itself here and do the formatting in the presenter, but okay shipping this as is

levimeahan (approved):

PR #658, opened 2024-04-05
Update battle.php - loot retreat fix

Voldrii:
fix so that if a player retreats, the winner receives half of the retreating player's loot, and the loser doesn't receive any of the winner's loot, only loses loot if they have any

### DIFF
--- a/pages/battle.php
+++ b/pages/battle.php
@@ -173,7 +173,7 @@ function processBattleFightEnd(BattleManager|BattleManagerV2 $battle, User $play
         $result .= "You have earned $village_point_gain point for your village.[br]";
 
         if ($battle->is_retreat) {
-            // Calculate rep gains
+            // Calculate rep gains and handle loot
             if ($player->reputation->canGain(UserReputation::ACTIVITY_TYPE_PVP) && UserReputation::PVP_REP_ENABLED) {
                 $rep_gained = $player->reputation->handlePvPWin($player, $battle->opponent, true);
                 if ($rep_gained > 0) {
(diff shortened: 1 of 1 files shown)


pages/battle.php
```diff
@@ -184,12 +184,14 @@ function processBattleFightEnd(BattleManager|BattleManagerV2 $battle, User $play
...
-            $system->db->query("UPDATE `loot` SET 
-                `user_id` = {$player->user_id}, 
-                `battle_id` = NULL 
-              WHERE `battle_id` = {$player->battle_id} LIMIT " . ceil($battle->getTotalLoot() / 2));
+    
+            // Calculate the half of total loot
+            $half_loot = ceil($battle->getTotalLoot() / 2);
+    
+            // Update the loot table for the retreating player to lose half of their loot
+            $system->db->query("UPDATE `loot` SET `user_id` = {$battle->opponent->user_id}, `battle_id` = NULL WHERE `battle_id` = {$player->battle_id} LIMIT $half_loot");
```
levimeahan, line 192: This query above looks like it will still assign half of the total loot from the battle to the retreating player, rather than half of the retreating player's loot. 

I think we needs to handle this in two steps
- Retreating player claims half their own loot, clearing the Battle ID from it
- Winning player claims all remaining loot for this battle (will be all their own loot + half the retreating players)

The retreating player's code branch will run first, so here in the winning branch we can just make this update all remaining loot to belong to the winning player and clear the battle ID from it.

pages/battle.php
```diff
@@ -235,18 +237,19 @@ function processBattleFightEnd(BattleManager|BattleManagerV2 $battle, User $play
...
             if (UserReputation::PVP_REP_ENABLED) {
                 $rep_lost = $player->reputation->handlePvPLoss($player, $battle->opponent, true);
                 if ($rep_lost > 0) {
                     $result .= "You have lost $rep_lost village reputation.[br]";
                 }
-                // Loot - winner takes half loser's if retreat
-                $system->db->query("UPDATE `loot` SET `user_id` = {$player->user_id}, `battle_id` = NULL WHERE `battle_id` = {$player->battle_id} LIMIT " . ceil($battle->getTotalLoot() / 2));
+                // Winner takes half of loser's loot if retreat
+                $half_loot_lost = ceil($battle->getTotalLoot() / 2);
+                $system->db->query("UPDATE `loot` SET `user_id` = {$battle->opponent->user_id}, `battle_id` = NULL WHERE `battle_id` = {$player->battle_id} LIMIT $half_loot_lost");
```
levimeahan, line 250: Per above, as this runs first we can use this to let the retreating player claim half of their own loot
```suggestion
                // Keep half of own loot, the rest will default to the winner
                $system->db->query("
                  UPDATE `loot` SET `battle_id` = NULL 
                  WHERE `battle_id` = {$player->battle_id} AND `user_id`={$player->user_id} 
                  LIMIT  (SELECT COUNT(*) FROM `loot` WHERE `user_id`={$player->user_id}) * 0.5");
```

PR #662, opened 2024-04-12
Fix regen cron not using max health

aumu1031:


cron_jobs/region_regen_cron.php
```diff
@@ -120,7 +120,7 @@ function processRegionRegenInterval(System $system, $debug = true): void
                 case 'castle':
                     // increase health, cap at max
                     $regen = (WarManager::BASE_CASTLE_REGEN_PER_MINUTE * WarManager::REGEN_INTERVAL_MINUTES) * max((1 + ($region_location['stability'] / 100)), 0);
-                    $region_location['health'] = min($region_location['health'] + $regen, WarManager::BASE_CASTLE_HEALTH);
+                    $region_location['health'] = min($region_location['health'] + $regen, WarManager($system, $region_location, $villages[$region_location['occupying_village_id']]));
```
levimeahan, line 123: Hmm, is this
```php
WarManager($system, ...
```
supposed to be calling a function?

levimeahan:
Addressed in 71d8ea3c009a6962a5c8842d109b9dfb0d196fc3 and follow-up 7c5752bed8c37666fcbc0963b085fae313618903

